PR #1, opened 2016-04-14
Simply android build environment

smartin015:


toddmedema:
1 - needs a note that building Android + iOS requires Mac / Linux
2 - getting build errors on my laptop

smartin015:
Added a note to the readme - build errors I'll fix after the landing, as it's working fine for me.

PR #4, opened 2016-04-21
A couple of independent changes

smartin015:
Reduced debounce amount, fixed some UI shiz with the Next button, updated build info in the readme, updated icons, and streamlined end-of-combat text.


toddmedema:
Looks good, +1

app/scripts/app.js
```diff
@@ -11,6 +11,27 @@
...
     this.isMobile = this.isAndroid || this.isIos;
+    document.addEventListener("backbutton", app.onBackKeyDown, false);
+  };
+  app.onBackKeyDown = function() {
+    // This is neither elegant nor efficient, but it works well. When native
+    // back button is pressed, loop through all expedition-card elements and 
+    // manually trigger a "return" event from the first expedition-card found
+    // that's "visible".
+
+    // Using querySelectorAll globally like this doesn't work on mobile (returns empty list)
```
toddmedema: @smartin015 My only concern here is that since Android is fragmented, it might work on your Android device but not others (as we've already been seeing with the white screen / app failing to load). I would at least want this solution to work on mobile web as extra peace of mind, but would ideally like this checked on a variety of Android devices (esp some of the ones that're currently failing to load, since there's definitely something different about them) before we fully close this.

If you're focused on coding, feel free to send me the APK for this build and I'll side load it onto every table in the apartment / test it

toddmedema:
Update: tested on several additional Android devices. All devices that loaded were able to use the back button as expected. Since the load failure bug is a separate issue, this PR looks good.

@smartin015 can you handle the merge conflicts, then feel free to merge

PR #6, opened 2016-04-23
Upgrade UI elements to be paper-like

smartin015:
Changes us using <a> tags to explicit <expedition-button> tags in order to have more explicit control over the styling of the button (hooray, no more :after CSS shenanigans!) There's also some quick style fixes to prevent buttons on the number-picker and expedition-picker elements from jumping around - they're now flex aligned.

Quest syntax remains unchanged (still uses "<choice>" for choices).


smartin015:
...these aren't the correct changes for some reason.

PR #10, opened 2016-04-24
Promote buttons to first-class polymer elements

smartin015:
This is necessary for finer control over button styling (flexboxes!) and also "ink" on-tap effects common to the paper-* elements built in to polymer.


toddmedema:
LGTM

PR #15, opened 2016-04-26
Ios fixes

toddmedema:
Note that this is currently rolling out to iOS beta testers as v1.0.8, which includes the config and splash screen PR's as well

Fixes a Safari rendering bug with the recent upgrade to expedition buttons:
screen shot 2016-04-26 at 12 14 56 pm


toddmedema:
@smartin015 fixed the Android build bug (turns out just making it lower case was enough, and that works on iOS too - no fancy sed command required). Also added the Android build script, couldn't think of any reason not to include in the repo

smartin015:
Don't submit this yet, still fixing major broken paths for icons.

toddmedema:
Closing; worked into https://github.com/Fabricate-IO/expedition-app/pull/120

PR #21, opened 2016-05-03
Updated mistress_malaise.txt

wls:
Removed comma (an invalid character) from an element's attribute list.


toddmedema:
Great catch @wls! And congrats on being the first person outside of the core team to contribute :)

PR #23, opened 2016-05-07
Allow clicking in browser

wls:
Added on-click elements for each on-tap element; added click event listener for checkbox. This mostly solves the problem of being able to test quest paths (not necessarily using the full app) via the browser (when the user-agent is set to a smart phone).

_I tend to do quest editing on my desktop and wanted to be able to test paths within my browser. The codebase didn't easily allow for that, even after switching my user-agent to a mobile device. This update allows clicking with a mouse from within a browser. It's enough to wander around a quest._


toddmedema:
Hey WLS! This is definitely something worth adding, since we're also getting requests for the app to work on desktop browsers (which will also require some re-styling, but listening for clicks is another blocking feature)

Two thoughts:

1. Having separate but identical "on-click" and "on-tap" handlers seems like it could lead to some bugs and maintenance headaches in the future. Wonder if there's a way to specify once, ie have on-click listen for both taps and clicks?
2. Following up on your browser remark / background statement - that's a bit unusual. Are you using Chrome and the Chrome developer's console to simulate being a mobile device? We haven't had any issues with taps being registered in that case

wls:
Admittedly, I made this patch for myself -- what I haven't been able to figure out is how you folks are testing quest files before releasing them.  The change-set wasn't intended for making the game playable in browsers, although it certainly leans in that direction (hence sharing the code).

When I fire up the app using **gulp serve**, _for me_ Safari starts up.  I tend to use Safari and Firefox (each provides slightly different diagnostics) and neither were responding to Polymer's `on-tap` event, despite the documentation: "[Use on-tap rather than on-click for an event that fires consistently across both touch (mobile) and click (desktop) devices.](https://www.polymer-project.org/1.0/docs/devguide/events.html)".  The addition of `on-click`, while ugly, got both of these environments working for me.

I suspect the problem is that I'm altering the User-Agent string, not that I'm emulating a mobile device ...or its custom events. (The User-Agent was to pacify the code because I wanted to work _in a browser as a browser_.)

I _strongly_ concur that the redundant event hooks feels ugly -- the "correct" way is the `on-tap`, which should handle both, but currently it isn't working.

The problem I'm actually trying to investigate on my browser is two fold:  1) I get a `cordova.js` resource unavailable, it is trying to load directly from the root directory.  2) I get an error that `.svg` (no filename) is also an unavailable resource at startup.

toddmedema:
@wls we're still working on testing - we have some structure laid out for unit testing, but it's not quite finished setting up yet.

Sounds like Safari is your default browser - in the interim / until we officially add Safari support, you should be able to open `http://localhost:5000/` in Chrome and use the dev tools there (which has the full device emulator that we've been using for testing)

Both of those errors are non-issues as far as we've been able to tell, just weird artifacts from the Polymer project. We see them on our browsers as well but haven't run into any issues.

wls:
**I'm canceling this pull request at the moment** -- while this fixes _other_ browsers, I've found an edge case that causes it to double-click (one for click one for tap) in Chrome, which only allows one to have an even number of adventurers.  In agreement with @toddmedema, I'd like to find a better way to solve the problem than to have two event hooks.

Issue #24, opened 2016-05-08
Label featured quests with time estimate


toddmedema:
Need to figure out a way to display time estimate in UI (ie 20-30min), and add that to the existing quests' metadata.

Initial guesses for timing are Albanus 20-40min and Malaise 40-60min

cc @smartin015

toddmedema:
Closing, duplicate of https://github.com/Fabricate-IO/expedition-app/issues/100

PR #26, opened 2016-05-08
Provide more documentation on the Quest Definition Language

wls:
Here's the stand alone change for the quest documentation.  It's just an additional file written in Markup that starts to go into a deeper dive about the language the quests are written in.

When I first started off looking at the sample quests, several things threw me.  One, it wasn't legal XML -- so I could use no automated tools to help create or validate the files.  Two, sometimes the structure appeared sequential, other times nested, and it took a little bit to figure out what was going on.  Three, some of the element names threw me -- for instance, roleplay _sounds_ like a verb or directive to the system and _choice win_ actually isn't a choice the user is making.

Ideally, I think revisiting the language would be of value, but for now I've found that with the following documentation, I and my friends have found understanding (and creating) quest files is much easier. This is the information I wish I had _before_ I ever looked at my first quest file; my hope is it will open the door even wider for quest contributions.

### DIFF
--- a/app/quests/quest_spec.md
+++ b/app/quests/quest_spec.md
@@ -112,7 +112,7 @@ Encounters _must_ have a win condition and a lose condition, but may not offer a
 ### Text Elements
 
 * **p** -- paragraph
-* **instruction** -- instructional text, not shown to advanced players
+* **instruction** -- instructional text, similar to paragraph, but styled differently to break the fourth wall and tell the players to do something, rather than the characters about the story plot
 * **b** -- bold
 * **i** -- italic
 
--- a/app/quests/quest_spec.md
+++ b/app/quests/quest_spec.md
@@ -81,11 +81,13 @@ Concludes the adventure, offering only the option to return to the menu to selet
 The end element behaves similar to the roleplay element, except it may only contain flavor text; that is, no choices, encounters, or further roleplaying.
 
 It has one mandatory attribute pick one:
-* **`win** -- this is a winning conclusion to the quest's end
+
+* **`win`** -- this is a winning conclusion to the quest's end
 * **`lose`** -- this is a losing conclusion to the quest's end
 _Omitting, even accidently, shows no text and ends the quest.
 
 And it has three optional attributes:
+
 * **`icon`** -- an image shown in the background of the text, used simply to provide visual flavor so text isn't visually boring
 * **`title`** -- a title shown for the page of text
 * **`id`** -- a symbolic label, used as the destination of a choice's goto
--- a/app/quests/quest_spec.md
+++ b/app/quests/quest_spec.md
@@ -1,0 +1,190 @@
+Quest Specification
+===================
+
+Quest files are used to build custom roleplaying scenarios with the Expedition App in a custom _Question Definition Language_.
+
+
+About the Syntax
+----------------
+
+While the files *look* like XML, they do not parse as valid XML. (Perhaps this will change in the future.)
+
+1. There is no single root element.
+2. Play usually follows sequential peer elements, although some content appears nested.
+3. Some parameters don't have values, and are assumed to be boolean true when present.
+
+
+### Counter Examples
+
+Not all of the files are quests.  At some point, these may be moved into another directory.
+
+You may find some examples of what _not_ to do
+in the following files:
+
+  * bad_goto.txt -- There's more than one element with the same `id`; `id`s should be unique.
+
+  * bad_malicious.txt -- A sample of JavaScript injection that performs an alert.
+
+
+### Tests
+
+There are other files that exist merely as tests. These, too, may be moved into another directory at some point.
+
+  * goto_test.txt - Performs a jump using a choice to another place in the question.
+
+### Quests as Examples
+
+Several quests exist, which additionally serve as examples of how to use the quest definition langauge.
+
+  * mistress_malaise.txt -- The undead have roamed across the land, your job is to stop them. *A fairly clean example of mostly linear story telling that utilizes most elements. There's a handful of side branches for variety.*
+
+  * oust_albanus.txt -- The town's cheese supply is being stolen, you need to find and stop the raiders. *Mostly linear content, but shows several nesting levels deep for choices.*
+
+  * ill_town.txt -- A mysterious illness has befallen a town.  *This adventure shows a new construct that's still being tested, called the `show-if` attribute on the `<choice>` element; it is used in conjuction with a new `<set-state>` element to carry game state between elments and make conditional content.*
+
+
+General Gaming Phase Structure
+------------------------------
+* Flavor text (`<roleplay>`) is presented to the players. Usually it contains blocks of plain HTML.
+* If there is no decision, a Next button is presented to the user and gameplay falls through to the next sequential element.
+* Decisions are offered with multiple choices (`<choice>`).
+* These often lead to an encounter (`<encounter>`) which features one or more enemies (`<e>`) that must be defeated. *The game handles the mechanic details.*
+* The encounter continues until it can resolve in only one of two ways, with a win (`<choice win>`) or a (`<choice lose>`).
+* Eventually the story line resolves (`<end>`), and the player is allowed to select a new quest.
+
+The quest is primarily a sequential path through the storyline, with nesting of elements within a choice to branch or the use of `goto`s to hop around in a non-linear fashion.
+
+
+Quest Elements
+--------------
+There are only a handful of elements.  Certain elements may have an `id` attribute, which must be unique across all elements.
+
+#### `<comment>`
+Anything in a comment element is ignored. Uses these for anotating quest files.
+
+#### `<roleplay>`
+This is not a verb, nor does it instruct the game to do anything. Rather it is for literal roleplaying: providing flavor text to set the context of the battles and move the story forward.
+
+The roleplay element may have three optional attributes:
+
+* **`icon`** -- an image shown in the background of the text, used simply to provide visual flavor so text isn't visually boring
+* **`title`** -- a title shown for the page of text
+* **`id`** -- a symbolic label, used as the destination of a choice's goto
+
+Within the roleplay element is simple HTML, primarily consisting of one or more paragraphs (`<p> ... </p>`).
+
+But it can also contain instructions to the players (`<instruction>`), as well as a list of choices (`<choice>`).
+
+#### `<end>`
+Concludes the adventure, offering only the option to return to the menu to selet another quest to play. Final rewards are doled out to the players.
+
+The end element behaves similar to the roleplay element, except it may only contain flavor text; that is, no choices, encounters, or further roleplaying.
+
+It also have three optional attributes:
+
+* **`icon`** -- an image shown in the background of the text, used simply to provide visual flavor so text isn't visually boring
+* **`title`** -- a title shown for the page of text
+* **`id`** -- a symbolic label, used as the destination of a choice's goto
+
+#### `<choice>`
+The choice element is, unfortunately, overloaded in purpose. The primary function is obvious: present an option to the players, and if selected, execute it.  This may either be to `goto` another element, or to perform a nested `<roleplay>` action, perform an `<encounter>`, or `<end>` the adventure story line.
+
+However, the choice element can also be used to detect win and lose conditions of an encounter.  Such elements contain _either_ the attribute `win` or `lose`.
+
+The choice element may also have a `goto` attribute, which is a reference to the `id` attribute in roleplay, encounter, or end elements.
+
+#### `<encounter>`
+Has the players battle one or more monsters, as specified by `<e>` elements.
+
+It also have two optional attributes:
+
+* **`icon`** -- an image shown in the background of the text, used simply to provide visual flavor so text isn't visually boring
+* **`id`** -- a symbolic label, used as the destination of a choice's goto
+
+Encounters _must_ have a win condition and a lose condition, but may not offer a choice. This gets particularly confusing as `<choice win>` and `<choice lose>` are mandatory, but `<choice>` without either attribute is illegal.
+
+
+Quest BNF
+---------
+It looks like the grammar for quests is still being refined and augmented. What follows is _not_ a formal BNF notation, but rather an attempt to capture general intent.
+
+```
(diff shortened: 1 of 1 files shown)


app/quests/quest_spec.md
```diff
@@ -0,0 +1,195 @@
...
+
+But it can also contain instructions to the players (`<instruction>`), as well as a list of choices (`<choice>`).
+
+#### `<end>`
+Concludes the adventure, offering only the option to return to the menu to selet another quest to play. Final rewards are doled out to the players.
+
+The end element behaves similar to the roleplay element, except it may only contain flavor text; that is, no choices, encounters, or further roleplaying.
+
+It has one mandatory attribute pick one:
+* **`win** -- this is a winning conclusion to the quest's end
```
toddmedema: missing a closing apostrophe

app/quests/quest_spec.md
```diff
@@ -0,0 +1,195 @@
...
+* **`lose`** -- this is a losing conclusion to the quest's end
+_Omitting, even accidently, shows no text and ends the quest.
```
toddmedema: Formatting on this line is weird

toddmedema:
@wls thank you for the thorough documentation! Left a couple of small notes re: the styling, but looks great.

Talked over the quest language with Scott, and we had an idea on the XML front, let me know if this feels like a good solution to you:

On one hand, we made some of the choices (like not having a root element, and not having ="true" for boolean elements, ie `<choice win="true">`) to make it easier to type when entering large quests.

On the other hand, if there are existing tools that make XML writing and validating easier (which we were not aware of when originally designing QDL), we certainly want to allow quest designers to use the best tools available.

As such, our proposed solution: a lightweight build script that converts lightweight into valid XML + HTML (valid HTML requires that all non-standard element names have a hyphen, so `<expedition-encounter>` instead of `<encounter>`). The app would then use the .xml script, but users could choose to write quests in the lightweight version as a .txt file, or write directly as the more verbose .xml version if desired.

Thoughts?

app/quests/quest_spec.md
wls: I've got a new pull request coming shortly. It should fix it. And more.

wls:
Several thoughts, and I _love_ the direction you're heading.

I'd strongly suggest making the document be [valid XML](http://www.w3schools.com/xml/xml_validator.asp), which means we could also publish a [DTD](http://www.w3.org/TR/html4/sgml/dtd.html) or [XML Schema](http://www.w3.org/standards/xml/schema) (XSD) that would allow lot of tools, including [free online editors](https://github.com/UNC-Libraries/jquery.xmleditor), to enforce correct syntax. It would make for slick [XLST](http://www.w3schools.com/xsl/default.asp) transformations.

That root element, which I recommend calling `<quest>`, is the _perfect_ place to address meta information (see issue #24) to have things like Quest Title, Synopsis, Optimum Number of Players, Estimated Duration, Credits, Version Number, Companion Resources, and so forth.

Attributes should _always_ have a value.  It is _okay_ for the XML _used by the game_ to be verbose and comprehensive, this allows for much better error checking, better debugging, and future proofing. As you wisely pointed out, there is nothing that prohibits producing a compiler or graphical environment that outputs the playable XML.

I'd also make the strong recommendation that `<choice win>` and `<choice lose>` should not be called 'choice' at all -- it isn't a choice, it's an _outcome_.  More over, it would be _much_ more clear if the elements were simply `<win>` and `<lose>`.

What the game desperately needs is the ability to set and test state, and not just true and false, but values, both numerical and text. Having conditional choices, flavor text, and encounters would make it possible to take less-linear paths _and solve puzzles_.

I noticed in `quests/ill_town.html`, there's a mechanism to kind-of do this (although the main branch doesn't seem to have backing logic for it).  Note that it's notation is `<set-state wrong-vial>true</set-state>` -- although this is _**not**_ the way to do it.  The reason is that the parser gets much more complex in having to examine the list of attributes and ignore known ones, which introduces spelling errors and typos as false positives. _Rather_, `<set value="true">symbol_name</set>` is trivial to parse, allows for _any_ value, and opens up a host of other attributes (such as conditional setting, calculated expressions, and lookup tables) for future extensions.

Also, I'm looking for a way to insert a symbol into flavor text, whether this be a Font-Awesome icon or one of the .SVG resources such as the d20 or the Range symbol. (To be fair, I haven't gone exploring for this yet, it just popped into my head.) I'd love to be able to inject some syntax like `{icon:name}` or `{svg:name}` and have the code do the right thing on presentation.

wls:
Just updated the pull-request with the latest changes at my end.  It fixes the format problem, addresses the `instruction` element, and adds some additional formatting elements for reference.

toddmedema:
Agree. At which point it'd also be good to have valid HTML for anyone who wants to use an HTML IDE (the only additional change would be including hyphens in all node names, such as `e-branch` or `q-branch`). Great stuff, thanks for bringing this up :)

win/lose: how about `<branch type="win|lose">`? Scott proposed `<branch>` instead of `<choice>` and `<outcome>` so that there's a single node type, which keeps the language as a whole simpler. Also, by keeping it as branch, it'd also support standard `<branch>` things like GOTOs, so that Encounters could potentially have custom outcomes (i.e. did you destroy the phylactery? jump to this other part of the quest)

variables: true. We'd also want the ability to increment and decrement, which starts making it more complex. One idea: we could utilize the math.js parser so that users could execute arbitrary math- and variable-related queries, such as `<set-state>vials = 0</set-state>`, `<set-state>vials++</set-state>`, `<set-state>gold += vials * 10</set-state>`, etc. which would open up a whole world of possibilities. http://mathjs.org/docs/expressions/parsing.html

variables: the other side of the equation (hah) is how variable logic will be shown to the user, ie if wrong vial = true, show this button. Since the math.js parser returns a value, we could potentially use that as well, i.e. `<button if="gold > 10">Buy potion</button>`

Instructions: it's actually not hidden from advanced users, since they could be advanced users but playing through this quest for this first time.

Also brings up a good point that we need to update how instruction blocks are displayed, dropped an issue: https://github.com/Fabricate-IO/expedition-app/issues/68

icons: Want to drop a separate issue with this? My main concern is that most common icon libraries aren't really appropriate for quests, i.e. they're mostly modern web icons.

wls:
**REGARDING FLOW CONTROL**
`<branch>` is more accurate from a perspective of _flow control_, and though I see what you're attempting to do is simplify the underlying constructs with a more generic operation (which I am _totally all for_), when I talked with my buddies who design multi-year campaigns for groups (uber-nerds), the consensus is that they want to think (and author) in terms of what the players will see. Quest designers appear to naturally think very declaratively and not procedurally; we should embrace that.

In other words, they _like_ the `<choice>` directive, because it _shows a choice_.  They'd rather have something that _indicates_ a win or lose state for resolution of the encounter.  Having that use the same element name is a high source of confusion and hurts new author adoption.  _I'm assuming the goal here is to get third parties writing quests supporting sales of the physical game._

The compiler-designer in me has no problem with this, because what the user provides as source does not have to be what is generated or executed under the hood.

_Where things "fall apart" for me is that it feels like DOM elements are being used as a data structure, and so the 'game model' is being force-fit into something that doesn't quite align nicely with intent. Having been down this road, such situations can paint one into a nasty corner where a grammar allows for "legal" use of generic instruction, but policy would dictate otherwise. This leads to over compensating with lots of edge-case checking passes, where a stricter grammar inherently forces the right syntax and intent._

_I am still trying to understand the 'why' of certain parts of the QuestParser._


**REGARDING VARIABLES**
I love the idea of using [MathJS](http://mathjs.org) to provide the ability to provide more expressiveness.  The only issue of concern is whether a malicious quest could execute code on the browser or spin a hosting server into a blocking loop.

That said, something in my gut says that there may be more value than just setting a value, but that some calculation of an expression may be of value (I can make puns too -- I saw what you did there with _equation_). If the element was `<evaluate>` then one might get dual use out of both setting state as well as retrieving it in a description.  `<p>You have <evaluate>gold</evaluate> pieces left.</p>`

That said, it feels more verbose than I'd like it. But the really cool part is that MathJS allows for multiple statements and key-value objects.  This. Would. Be. Amazing.

Being able to couple an `if=` attribute to an[y] element would provide a whole host of _replayable_ content.

**REGARDING `<INSTRUCTION>`**
Can you clarify what this directive is for?  From what we could tell, it looked like it operated nearly identically, from a visual perspective, to the `<p>` element, with maybe a pixel or two of indentation. When I quickly scanned the source, it looked like in `expedition-settings.html` that the `helpnone` option "Hides all instructional text."

I'm now taking that to mean that tutorial and what-to-do-now pages in the engine itself are not displayed, and it is not referring to the `<INSTRUCTION>` element at all.  That then begs the question, if the quest doesn't have its own optional instructions, what is the element used for and when is it used?

Based on this discussion alone, I'm taking #68 is actually intended as a _stylistic_ display allowing the game to [break the fourth wall](https://en.wikipedia.org/wiki/Fourth_wall) and tell the gaming players what to do as opposed to the characters they play what's happening in the story line.

**REGARDING ICONS**
In the `ill_town.html` quest, which _does not parse_, there are injections like this in the description:

```
<p>Grab (fa-coffee) and steal a (fa-book)</p>
```
Which, I take was intended to render as: Grab a ☕ and steal a 📖 

While I do _not_ like the syntax (as it'd be far too easy to confuse with parenthetical texts), the idea of injecting .SVG content is sound.  (And new thought: could also be used for sound effects!)

I concur most icon libraries are web-application based, although having a mechanism would invite font designers to make additional fantasy role playing font. Like [Rpg-Awesome](https://nagoshiashumari.github.io/Rpg-Awesome/) or [rpg-icon-font](http://opengameart.org/content/rpg-icon-font). Any chance of bringing these into the game?

I have at least one quest-writer who has figured out some new clever mechanics using the game resources in unusual ways would like to be able to _in the description_ direct someone to roll the d20 using the same graphical notation as the game does, for the sake of consistency.


RECOMMENDATION: Let's break this up into separate threads of thought.

smartin015:
Hey wls@, just wanted to introduce myself since I've spoken through Todd a couple times (he currently has a lot more time for working on Expedition than I do). Thanks for taking an interest in making Expedition the best game it can be, and for your awesome and insightful work so far.

I agree with the recommendation to break the topics up, as this thread's becoming a bit unwieldy. See #70 for variable discussion, #68 for the instruction tag, and #71 for icon injection. 

The `<instruction>` tag currently does basically nothing. You're right - it's a way to be out-of-character for a moment to discuss game mechanics, and it's confusingly the same as regular RP text. #68 is supposed to fix this.

ill_town.html is not an actual quest, and is more of a wishful thought of what could be. We made it before we did much work on the language. When you look at that one in particular, think more about the ideas and less about the implementation in the file ;)

Alright, onward to...

**Flow Control**

I understand your argument for clarity and strictness of grammar, especially since a `<choice>` node does slightly different things in the two cases (one is a user choice, one is a forced event by the encounter). In the short term it's absolutely the best way to go to split this into two tags, both for clarity and the mismatch of ability and policy that you mentioned. 

In the long term, however, we're planning enhancements such as #32. This one in particular adds user choice to an encounter, so `<choice>` tags would actually make sense as children of an `<encounter>` tag. Thenceforth, users will need to remember to specifically use `<outcome>` for win/lose states and `<choice>` for user choices, which seems non-ideal.

We developed this game with a strong bias towards simple rules/low mental overhead, and we'd like QDL to be the same experience. Simple interfaces are far easier to extend and reason about than feature-rich ones, which is why I'm leaning heavily towards using a single tag for both (at least until we have a better idea what the final form of the language will look like).

There's also a third option we haven't yet discussed: making win/lose a "goto"-style attribute of the encounter:

```
<encounter win="do1" lose="do2">
  <e>enemy1</e>
</encounter>
<roleplay id="do1>
  <p>players won the encounter!</p>
</roleplay>
<roleplay id="do2">
  <p>players lost the encounter...</p>
</roleplay>
```

Roleplay tags could still be embeddable within the encounter, if the user wants to keep the nested heirarchy. Users can also more easily have a generic "you die" message for lost battles by just jumping to an endgame tag for all losses that don't advance the story.

This separates the two use cases of `<choice>` into different explicit syntax, extends nicely in the case of #32, doesn't require any top-level additions to the language, and lessens the chance for a user to mistakenly put win logic (`<outcome win>` or `<choice win>`) in a place that's not an encounter.

Thoughts?

wls:
I absolutely _love_ the idea of a `win=` and `lose=` set of attributes -- this is much more clear in that it shows that win/lose conditions are a part of (and thus a consequence of) the encounter.

It also has the indirect impact of un-nesting encounter blocks, making the contents of the encounter element just about the encounter itself as well as making the overall quest far more readable.  It also is paves the way easily for the future when state is formally added so that adventures can easily reuse plot-paths without trying to reach inside the scope of an encounter.

My vote is to move forward with that notation.

wls:
I'd also like to propose an optional `retreat=` option as well, giving the adventures the opportunity to run away.  This has an impact on #32 -- which would allow one to decide _when_ such an option might become available during combat.

smartin015:
Let's move forward with win/lose attributes, then. 

Retreat could be cool as long as it's optional, since it suggests a pretty long remainder of play (unlike defeat, which is most commonly the short and simple "everyone dies and you lose").

toddmedema:
Awesome! `#teammwork`

@wls want to update the documentation in this PR to match the new flow, and we'll work on our end to update the code? (note that it might take us a few days / moving this weekend).

Thanks for all of the input!

toddmedema:
Actually - @wls @smartin015 maybe we don't want retreat as an attribute on the encounter:

Since retreat is an optional user interaction, it must be defined somewhere else - specifically, the "retreat" option would be shown in the custom combat prompt: https://github.com/Fabricate-IO/expedition-app/issues/32

Which means that, realistically, it doesn't have to be labelled retreat, and that as a `<choice>` node, it already has the functionality to `goto`. So, rather than limiting ourselves to `retreat` defined on the encounter, we can give quest writers unlimited flexibility via the `choice goto` inside of the custom combat prompts.

wls:
@toddmedema, before I update the documentation, did you want to commit it to document _this_ version?

toddmedema:
@wls agreed. Merging this in, we can create the revised documentation as a new PR that can pair with the matching code update

PR #72, opened 2016-05-10
Deal damage per round instead of per second

toddmedema:
closes https://github.com/Fabricate-IO/expedition-app/issues/69

### DIFF
--- a/app/elements/combat-timer-card.html
+++ b/app/elements/combat-timer-card.html
@@ -17,12 +17,11 @@
         this.active = false;
 
         this._roundTimeRemaining = 0;
-        this._roundTimer = 10000; // starting round time, varies with difficulty
+        this._roundTimeTotalMillis = 10000; // starting round time. TODO change based on difficulty
         this._roundDamage = 0;
         this._lastAttackMillis = 0;
         this._timerStartMillis = 0;
         this._lastRenderMillis = 0;
-        this._flashText = null;
         this._flashAlpha = 0;
         this._touches = [];
 
@@ -44,9 +43,8 @@
       start: function() {
         this.active = true;
         this._timerStartMillis = Date.now();
-        this._setFlashText("Attack!");
         this._lastRenderMillis = this._timerStartMillis;
-        this._roundTimeRemaining = this._roundTimer;
+        this._roundTimeRemaining = this._roundTimeTotalMillis;
         this._renderLoop();
       },
       stop: function() {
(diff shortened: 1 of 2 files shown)


app/elements/combat-timer-card.html
```diff
@@ -16,6 +16,8 @@
       ready: function() {
         this.active = false;
 
+        this._roundTimeRemaining = 0;
+        this._roundTimer = 10000; // starting round time, varies with difficulty
```
smartin015: If this varies with difficulty, why is this set to 10s here? Also, is this milliseconds?

Prefer this._roundTimerMillis

smartin015: Also looks like _roundTimer holds the total time for the round and is not in fact a timer, while roundTimeRemaining is a timer that ticks down to zero.

app/elements/combat-timer-card.html
```diff
@@ -44,27 +46,35 @@
...
         // We want to stop the timer using the number of *alive* adventurers.
         if (!this.globals.multitouch || this._touches.length >= this.adventurers) {
           this.stop();
         }
         e.preventDefault();
       },
-      _drawDamage: function() {
-        this.ctx.font = "120px MarkoOne";
+      _drawTimer: function() {
+        var tempTimer = this._roundTimeRemaining/1000;
```
smartin015: maybe formattedTimer? "temp" in variable names is discouraged.

app/scripts/combat_mechanics.js
```diff
@@ -136,15 +114,22 @@ Encounter.prototype.resetSurgeCounter = function() {
...
   s.expectedDamage = Math.ceil(s.turnTimeMillis * (this._stats.tierDmgRate) / 1000);
   s.nextSurgeRounds = this.getNextSurgeRounds(s.turnTimeMillis);
-  s.tier = this.tier;
-
+  s.tier = this.getTier();
+
+  // enemies each get to hit once - twice if the party took too long
+  s.roundDamage = 0;
+  for (var i = 0; i < s.tier; i++) {
+    s.roundDamage += this._randomAttackDamage();
```
smartin015: This can swing wildly - from 0 to 2*tier.

What about using tierDmgRate for calculation instead?

Also, would be nice to have the combat logic in a separate function from endRound.

toddmedema: Split out to separate function.

Hmm, what's bad about it being random? Since it's calculated as per-tier attacks, the odds of 4 tiers all criting (10% chance) in a single round is 0.1 ^ 4 = 0.0001

toddmedema:
@smartin015 fixed all comments. Also updated the timer screen to show a red background when players are over-time

smartin015:
Disregard randomness comment - thought hit/miss/crit was being weighted evenly.

PR #74, opened 2016-05-11
add href support to expedition-button

toddmedema:
closes https://github.com/Fabricate-IO/expedition-app/issues/73

closes https://github.com/Fabricate-IO/expedition-app/issues/37

### DIFF
--- a/app/elements/base/expedition-button.html
+++ b/app/elements/base/expedition-button.html
@@ -73,7 +73,18 @@
       is: 'expedition-button',
       properties: {},
       _onTap: function(e) {
-        this.fire('tap');
+        var target = e.target;
+        while (target.nodeName.toLowerCase() !== 'body' && target.nodeName.toLowerCase() !== 'expedition-button') {
+          target = target.parentNode;
+        }
+
+        if (target.getAttribute('href')) {
+          window.open(target.getAttribute('href'));
+        }
+        else {
+          this.fire('tap');
+        }
+
         e.stopPropagation();
       }
     });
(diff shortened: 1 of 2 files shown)


app/elements/quest-tester.html
```diff
@@ -87,6 +87,15 @@
       _openURLChooser: function() {
         this.$.urlchooser.open();
       },
+      _openBrowserDeviceEmulation: function() {
```
smartin015: This isn't very sustainable. Suggest adding an "href" attribute to the expedition-button template.

toddmedema:
@smartin015 good call, updated to the more sustainable system

app/elements/base/expedition-button.html
```diff
@@ -73,7 +73,18 @@
       is: 'expedition-button',
       properties: {},
       _onTap: function(e) {
-        this.fire('tap');
+        var target = e.target;
+        while (target.nodeName.toLowerCase() !== 'body' && target.nodeName.toLowerCase() !== 'expedition-button') {
+          target = target.parentNode;
+        }
+
+        if (target.getAttribute('href')) {
```
smartin015: Interestingly, this'll open an href on the <body> element.

toddmedema: @smartin015 yep - even better, realized that we can just check for expedition button - the alternative is that it keeps looping all the way up to the root HTML element, and then parentNode is undefined. Check out the new push

toddmedema:
removing not ready to merge label, ready for final review

smartin015:
LGTM

PR #77, opened 2016-05-17
Add CONTRIBUTING.md

toddmedema:
Based on the Hapi.js contributing guide: https://github.com/hapijs/hapi/edit/master/CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,28 @@
+# How to contribute
+
+We welcome contributions from the community and are pleased to have them.  Please follow this guide when logging issues or making code changes.
+
+## Logging Issues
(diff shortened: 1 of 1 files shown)


smartin015:
Is there some way we can make this consistent with the one for expedition-cards? Maybe just link to that contrib guide from here? No need to have duplicate documentation.

toddmedema:
Hmm. Well, it's unlikely that this will change much, and it includes repo-specific links (ie to the new issues page), so I'd be inclined to keep them both for simplicity.

smartin015:
Could always include separate links in the single contrib file... but as long as you're fine keeping both of these docs in sync, SG.

PR #80, opened 2016-05-23
Fixed encounter element

wls:
The encounter element can not have flavor text. Fixed the BNF grammar to reflect the correction.


wls:
Discovered that the documentation had a minor error in the grammar that incorrectly stated encounter tags could have roleplay-like text on it.  They can't.

smartin015:
Thanks! And sorry for the delay - moving houses and unpacking takes forever.

PR #83, opened 2016-06-25
Improved quest parser errors & updated encounter config

smartin015:
The "invalid enemy" error checking in combat-card overwrote the name before printing it >.<

We also significantly adjusted health values for enemies since the last config pull.


toddmedema:
+1

Issue #52, opened 2016-05-08
Limit adventurer count in combat section


smartin015:
Shouldn't be able to increase adventurers past the value in settings. If the user tries to do so, show a toast or other indicator that they're being dumb and should do it in settings specifically.

toddmedema:
Closed by https://github.com/Fabricate-IO/expedition-app/pull/84

Issue #38, opened 2016-05-08
When # of adventuers at min or max, don't show that arrow (ie can't go below 0)


smartin015:
In settings, combat, and enemy tier sum in combat.

toddmedema:
Closed by https://github.com/Fabricate-IO/expedition-app/pull/84

Issue #39, opened 2016-05-08
Limit # of adventurers between 1 and 6 inclusive


smartin015:
In settings and combat.

toddmedema:
Closed by https://github.com/Fabricate-IO/expedition-app/pull/84

PR #84, opened 2016-06-25
Limits and arrow hiding on number picker element

smartin015:
Added min/max limits to number-picker and restricted adventurer count. Also hid arrows when at min/max


smartin015:
Fixes #38 and #39

smartin015:
Also fixes #52

toddmedema:
+1

PR #88, opened 2016-06-25
Fixed instruction tag in quest parser and fixed malaise instruction tags.

smartin015:
Instruction tag didn't initially allow for paragraph separation of instructions.


toddmedema:
+1

PR #89, opened 2016-06-25
Update watermark alignment to move on scroll, attach to bottom.

### LINKED ISSUE #45: Adjust watermarks to not float
"attach" to the bottom right card corner. Set cards' min height (in css or js) to be (slightly less than) view height, so that even on small pages, bottom corner is at screen bottom.

smartin015:
Closes #45


toddmedema:
+1

Issue #82, opened 2016-06-24
Turn help setting into difficulty setting


toddmedema:
Options: Tutorial, Easy, Normal, Hard, Impossible

Tutorial difficulty also enables full help
Easy difficulty also enables some help

As far as difficulty, adjusts damage here: https://github.com/Fabricate-IO/expedition-app/blob/master/app/scripts/combat_mechanics.js#L91-L103

Currently difficulty is normal; knobs we can twiddle:
- # of attacks that occur (ie currently 1 attack per tier, but on tutorial could only be 0.5 attacks per tier)
- timer duration (ie how long before extra damage occurs. Currently 10s, on tutorial could be 60s)
- extra damage ratio (ie amount of extra damage that occurs if timer goes over. Currently 2x, could be 1.25x on tutorial)

PR #96, opened 2016-08-13
Change helplevel to difficulty setting

smartin015:
The help settings are now included inside of globals.difficultySettings. The difficultySettings object is passed to the combat timer (on start to set the round duration) and to combat_mechanics (on round end to scale the damage dealt).


smartin015:
Closes #82

toddmedema:
+1

PR #98, opened 2016-08-13
Back button now exits the app. Closes #19

### LINKED ISSUE #19: Pressing "Back" from home screen should close app
Following the standard UX for Android, pressing the hardware "BACK" button while on the homescreen of the app should close the app. It should ideally kill the app process, but even just minimizing the app would be a good start.

smartin015:
Closes #19


toddmedema:
+1

PR #109, opened 2016-08-14
Massive update to quest design language (QDL) and update to existing quests/docs

### LINKED ISSUE #75: Update QDL encounter system
Inside of the encounter, can have `<roleplaying>` nodes with an `on` attribute that acts as an event handler. In the case that win / lose happens but there's no handler, it should hop out of the loop and continue going.

on="win"
on="lose"
on="round"

note that these roleplaying cards should still have a black background

in the future, could do things like on surge, player death, etc

also, change `<end>` from being a clone of roleplaying, to just being an `<end></end>` that you can stick at the end of any roleplaying block

Will need to update core quests + quest spec

https://github.com/Fabricate-IO/expedition-app/pull/26

### LINKED ISSUE #92: Renaming encounter events to combat
Both in the UI and in the quest system

### LINKED ISSUE #100: Add root element to quest language
Add a root `<quest>` element with metadata (time estimate, title, synopsis, number of players, credits, version number, companion resources)

### LINKED ISSUE #101: Require legal XML for quest spec and offer a translator.
Store final version in full valid HTML/XML and offer a means to translate into that from the existing lightweight version.

### LINKED ISSUE #102: Fix existing quests to work with QDL updates

### LINKED ISSUE #104: Update / improve QDL documentation, including how to use the new tooling

### LINKED ISSUE #106: Move choice text into attribute of choice
Ie instead of
`<choice>Blah blah<roleplaying>something happens</roleplaying></choice>`
it becomes
`<choice text="Blah blah" icon="sword"><roleplaying>something happens</roleplaying></choice>`
which flows better and more intuitively matches the fact that that text will be inside of a button / not a whole separate element on the page.

Question: how would embedable icons be handled in the future? Should be fine to include the `<icon class="sword"></icon>` in the text? Though might get hairy with quotes.

Possible implementation: have a separate attribute for `text` and `icon`. See this material ui library for a nice implementation of this: http://www.material-ui.com/#/components/list

smartin015:
A few changes:

- `<encounter>` is now `<combat>`
- `<encounter>` triggers `<event>s` which have similar action to `<choice>s`
- Quests end with an `<end></end>` tag with no children and no attributes
- All quests now have a `<quest>` root with metadata
- All quests are now valid XML

Note that while IF and OPERATION are listed in the language definition, they are not yet implemented.

closes #102 closes #101 closes #100 closes #92 closes #75 closes #106 closes #104


toddmedema:
+1

@wls check this out! Took a while, but we finally were able to make many of the QDL improvements we'd talked about :)

PR #113, opened 2016-08-14
Added operations and conditional logic to QDL. Also checking in test …

### LINKED ISSUE #70: Op nodes & if attribute in Quests
Use [MathJS](http://mathjs.org/) to provide the ability to provide more expressiveness in quests via `<op>x=2</op>` blocks and `if=(math expression such as x > 1)` attributes on all blocks

If `if` evaluates to true, that block is executed. If it is a type of block that there can only be one of (ie an on=win), any on=win following it are ignored.

For example:

op: x = 2
encounter
event on=win if=x>2: skip me!
event on=win if=x>1: run me!
event on=win if=x>0: skip me!

Note that we should also recommend in the docs that anywhere if's are used, we recommend having a counter statement. For example if you have a `win if=demonSword`, you should also make sure to have a `win if=!demonSword`, or a default / catchall plain `win` node

smartin015:
…quests and helper icon

closes #70


toddmedema:
+1

PR #116, opened 2016-08-18
Markdown QDL, XML schema validation, and changes to QDL language.

smartin015:
So this one may have gotten away from me a little :P enjoy the small novel.

- Created a markdown version of QDL and converted the two existing quests to it (`<op>` doesn't yet work!)
- Created a markdown parser/converter to turn the original *.md file into *.xml file containing QDL.
- Crafted an XML schema that can be used to validate QDL in XML format
- Updated documentation for the above
- Changed `<end>` tags to `<trigger>` tags in preparation for future actions (like going back into combat)
- Small initial work into keeping game state available for quest `<op>` tags.


app/elements/quest-card.html
```diff
@@ -101,6 +102,8 @@
         this._isWin = undefined;
 
         this._parser = new questParser();
+        this._parser.setGameState(this.globals);
```
smartin015: All mentions of GameState are for later work getting game state (# alive enemies, encounter tier, number of players, etc) into `<op>` codes.

app/quests/build/mistress_malaise.xml
```diff
@@ -0,0 +1,313 @@
+<quest title="Mistress Malaise" summary="Mystery, Misfortune, and a Mistress."
```
smartin015: This and the Albanus quest were generated from the markdown files.

toddmedema:
Nice work! Thoughts for testing?

smartin015:
NONE >:{

But actually, about to begin work on a translator for the other direction. Can end-to-end test with that, then have a couple golden output files to compare against specific examples.

toddmedema:
Gotcha

+1

PR #117, opened 2016-08-20
Added bidirectional quest translation

smartin015:
We can now convert the albanus quest from mardown to xml and back with zero diffs.

The to_xml file is a much cleaner merger of all the old crufty md->xml parser files, to the point where it's essentially new.


toddmedema:
Question: do we still want this code to exist in the app, or should we clear it out of the app and move all of it into the new quest builder that you knocked out over the weekend?

PR #130, opened 2016-11-01
Convert all styles to SASS, removing old Dialogs code

smartin015:
Yep, turns out using sass was about as simple as "s/less/sass/g" in webpack.


app/style.scss
```diff
@@ -0,0 +1,475 @@
...
+  font-family: MinionPro;
+}
+
+.inline_icon {
+  width: 5vw;
+  margin-bottom: -0.5vw;
+}
+
+.base_button {
+  &.dark > button {
```
toddmedema: does this need to be a direct descendant? Seems like all buttons on a dark page, even if they're inside another div, should be darkened?

app/style.scss
```diff
@@ -0,0 +1,475 @@
...
+    background-color: $bg_dark_interactive !important;
+  }
+  &.disabled > button {
```
toddmedema: is .disabled not applied directly to the button? Also, since this is nested inside of .base_button, seems a bit redundant / slower to then look back up to see if it's inside a button?

app/style.scss
```diff
@@ -0,0 +1,475 @@
...
+  padding: 0;
+  font-family: MinionPro;
+}
+
+.inline_icon {
+  width: 5vw;
+  margin-bottom: -0.5vw;
+}
+
+.base_button {
```
toddmedema: any reason we need a special class for `.base_button` instead of just selecting all buttons?

toddmedema: Ah, based on the fact that there's also a `> button` selector with styles below, I think I see what's going on now. There's some extra styling going on around / on top of the actual button, that couldn't be done with a single element. #TheJoysOfHTMLCSS

app/style.scss
```diff
@@ -0,0 +1,475 @@
...
+  &.dark {
+    color: white;
+    background-color: black;
+  }
+  width: 100%;
+  height: 100%;
+  background-color: inherit;
+  color: inherit;
+  .value {
+    font-size: 110px;
```
toddmedema: Magic number? In cases that we're scaling the rest of the UI, this'll break (as will margin-top below on L354, which can also be written as `-VAR/2`

app/style.scss
```diff
@@ -0,0 +1,475 @@
...
+        bottom: $hhuge;
+        margin-right: -10%;
+        margin-top: -25%;
+        opacity: 0.2;
+      }
+    }
+    .scrollbox_bottom {
+      position: absolute;
+      bottom: -6vw;
+      right: -6vw;
```
toddmedema: Magic numbers!

app/style.scss
```diff
@@ -0,0 +1,475 @@
...
+    }
+    .scrollbox {
+      position: relative;
+      display: flex;
+      flex-direction: column;
+      width: 100%;
+      > img {
+        position: absolute;
+        height: 36vh;
+        width: 36vh;
```
toddmedema: Magic numbers!

app/style.scss
```diff
@@ -0,0 +1,475 @@
...
+  width: 100%;
+  height: 100%;
+  margin: 0;
+  padding: 0;
+  font-family: MinionPro;
+}
+
+.inline_icon {
+  width: 5vw;
+  margin-bottom: -0.5vw;
```
toddmedema: Magic numbers!
We can probably say that `margin-bottom: -WIDTH_VAR/10` for now

app/style.scss
smartin015: Yep, exactly that :)

app/style.scss
smartin015: Good point - this also simplifies the prop interface to these base components. Also found a couple manually constructed styles that could go away too.

app/style.scss
smartin015: Yep, you're right. Fixed.

app/style.scss
smartin015: Confound these magic numbers... fixed.

app/style.scss
smartin015: Fix'd

app/style.scss
smartin015: Fix'd

smartin015:
Manually merged.

PR #141, opened 2016-11-29
Buildfix & bugfix

toddmedema:
(web app deployed)

Done:
- fixes relative image bug in apps
- fixes web app, Android, iOS auth

Still to do:
- fix iOS API calls


toddmedema:
@smartin015 note that the fix for the web app was to dynamically load the google auth javascript, instead of statically include it. Apparently, because that JS lib is still in beta, it's been known to change and break every few months

smartin015:
I'd suggest running this on a phone with wifi/data turned off to ensure it works. Otherwise, LGTM.

toddmedema:
Confirmed on phone on airplane mode

toddmedema:
Well, that is to say, confirmed that the offline parts of the app didn't break, but the online quest section obviously didn't work, since even if they could authenticate, they wouldn't be able to ping the API

toddmedema:
Latest commit fixes iOS auth / allows you to log in, but now when you try to search, nothing happens...

smartin015:
LGTM.

PR #143, opened 2016-12-03
Add op support for roleplay and combat cards.

smartin015:
We now pass a context along as we explore quest nodes. This context lets us perform operations (ops) via double-curly "{{ test }}" mathJS operations inside roleplay cards. 

Showing/hiding based on "if" attributes is also now supported.

Also setup unit testing!


smartin015:
Confirmed to work within the IDE :champagne:

PR #147, opened 2016-12-06
Fix malaise quest + goto next buttons

toddmedema:
Closes https://github.com/Fabricate-IO/expedition-quest-ide/issues/57


toddmedema:
@smartin015 note that this means the currently published Malaise quest is broken; since it's published under your account, could you unpublish it when you have the chance? (renamed it to Mistress Malaise - old in the drive folder). If we want to, we could then publish the new version, though that doesn't seem strictly necessary / seems a bit redundant given it's featured

toddmedema:
Note: merging so that I can keep making progress, but should still review post-merge

smartin015:
LG

PR #148, opened 2016-12-07
fix instructions; allow multiple per card

toddmedema:


toddmedema:
merging to keep moving; should review later

app/QuestParser.tsx
```diff
@@ -331,19 +331,15 @@ export function loadRoleplayNode(node: XMLElement, ctx: QuestContext): RoleplayR
...
-  var instruction: JSX.Element = null;
+  var instructions: Instruction[] = [];
 
   var newScope = Object.assign({}, ctx.scope);
 
-  // Keep track of the number of choice nodes seen, so we can
-  // select a choice without worrying about the state of the quest scope.
-  var idx = -1;
+  var choiceCount = -1;
+  var instructionCount = -1;
```
smartin015: choiceCount is used later to reference users clicking on specific choices... it looks like instructionCount is only to add a key attribute in the react render. Could we remove this count and instead do a programmatic enumeration inside the Callout render function?

toddmedema: Just double checked the Quest Parser, looks like choiceCount is also only pass along to the roleplaying component... seems like we could also programmatically enumerate choiceCount.

Dropping a separate issue to tackle, since that can be roped into adding more tests: https://github.com/Fabricate-IO/expedition-app/issues/151

PR #149, opened 2016-12-07
fix conditional combat check

toddmedema:
Closes https://github.com/Fabricate-IO/expedition-quest-ide/issues/58

Merging now; should be reviewed later


app/QuestParser.tsx
```diff
@@ -168,8 +167,14 @@ export function loadCombatNode(node: XMLElement, ctx: QuestContext): CombatResul
         }
         break;
       case 'event':
-        winEventCount += (c.attr('on') === 'win' && _isEnabled(c, newScope)) ? 1 : 0;
-        loseEventCount += (c.attr('on') === 'lose' && _isEnabled(c, newScope)) ? 1 : 0;
+        switch (c.attr('on')) {
```
smartin015: This removes _isEnabled checking... which means we're now open to generic runtime exceptions e.g. if a user conditionally disables all "on win" conditions... but really, we're only solving the smaller problem here.

Opening a new bug to add correct conditional checking across all elements.

toddmedema: From in person: actually not needed, there's another (redundant) check earlier in the loop

PR #150, opened 2016-12-07
fix conditional choice bug

toddmedema:
Was triggering the incorrect "choice" node because, when rendered, the index was determined by incrementing only on contextually valid choices - but, when clicking, it incremented the counter for non-valid choices as well


toddmedema:
Merging for progress; should review later

smartin015:
We should probably unit test this...

toddmedema:
@smartin015 yep, added an issue so that we don't forget: https://github.com/Fabricate-IO/expedition-app/issues/151

PR #155, opened 2016-12-10
fix search

toddmedema:
Rather than make "in last X time" be dependent on device's timezone, have it just pass the offset to the server, and let SQL calculate.

Merging for progress, but still worth reviewing


smartin015:
Makes sense to me - nice cleanup!

PR #159, opened 2016-12-20
Fix SVG coloring in quest creator

toddmedema:
Closes https://github.com/Fabricate-IO/expedition-quest-creator/issues/71


smartin015:
LGTM

PR #162, opened 2016-12-29
allow instructions to override default icon

toddmedema:
Once this is merged in, we'll need to update the quest creator to point to the new commit


app/components/Roleplay.tsx
```diff
@@ -29,8 +29,13 @@ const Roleplay = (props: RoleplayProps): JSX.Element => {
   });
 
   var instructions: JSX.Element[] = props.roleplay.instructions.map(function(instruction: Instruction): JSX.Element {
+    let icon = 'adventurer';
+    // if there's an icon at the begining, don't show default icon
+    if (instruction.text.indexOf('<p><img') === 0) {
+      icon = null;
```
smartin015: It seems weird to pass the icon via innerHTML instead of by the icon attribute... and it may break if we apply additional styles to the passed icon. Could we instead extract the image path from the element and pass it into the icon attribute?

toddmedema: @smartin015 good call, new PR (will wait to merge for your feedback: https://github.com/Fabricate-IO/expedition-app/pull/165)

danielwiktorczyk:
Very cool! The instruction override > [icon] is working for me!

PR #164, opened 2016-12-29
add new icons

toddmedema:
- caduceus
- female
- male
- anticlockwise
- clockwise
- key


smartin015:
From what I could see in the repo (the icons show up as SVG code in the commit, and I wasn't keen to track them all down) these don't match the style of our current icon set. We should open these up in illustrator and apply the expedition brush settings to make them feel more like the existing icons.

toddmedema:
Yep, that's correct. Currently have them labeled as "not Expedition themed" in the icon docs. Feel free to submit a pr that improves their style :-)

smartin015:
Added #167 to track.

PR #166, opened 2016-12-29
adjust damage based on number of players

### LINKED ISSUE #65: Balance the game for different numbers of players
Blocked by #69 

Factors to consider:
- loot (the way we do it now, fixed loot pile, actually works well)
- level scaling (ie that larger parties become stronger faster)
- that larger parties have more support abilities, and support abilities are also more powerful
- that being said, giving out the same amount of loot for large and small parties isn't a great UX - small parties would get overwhelmed with loot, large parties not all players get loot

Perhaps take the current damage output as being considered well-balanced for 4 players and divide and multiply from there. I suspect that damage won't need to scale dramatically for varied #'s of players - I would propose:
1 - 0.25 (though this might end up being a special case anyway)
2 - 0.5
3 - 0.75
4 - 1
5 - 1.25
6 - 1.5 (basically forcing at least one person to always be support - might also need to reduce the potency of support in large parties, ie not "enemies deal no damage" but "enemies deal 4 less damage" so that damage multiplier still has an effect)

For actually dealing the damage, rounding is going to be a problem (2/.25 = .5, which will always round to 1, instead of averaging to .5). As such, we can use the fraction as the probability of Math.floor vs Math.ceil. (ie .4 has a 40% chance of ceil, 60% chance of floor)

toddmedema:
closes #65


app/reducers/combat.tsx
```diff
@@ -40,21 +41,29 @@ export function isSurgeRound(combat: CombatState): boolean {
   return (surgePd - (rounds % surgePd + 1)) === 0;
 }
 
-export function generateCombatAttack(combat: CombatState, elapsedMillis: number): CombatAttack {
+export function generateCombatAttack(combat: CombatState, elapsedMillis: number, settings: SettingsType): CombatAttack {
```
smartin015: We should unit test this at some point...

app/reducers/combat.tsx
```diff
@@ -40,21 +41,29 @@ export function isSurgeRound(combat: CombatState): boolean {
...
   // Attack once for each tier
+  let damage = 0;
   for (var i = 0; i < attackCount; i++) {
     damage += _randomAttackDamage();
   }
 
-  // Scale according to multiplier, then round to whole number.
-  damage = Math.round(damage * combat.damageMultiplier);
+  // Scale according to multipliers, then round to whole number.
+  // For small number of players and minimal damage, negate multiplier to prevent always 0 damage
```
toddmedema: Hmm, should do the opposite - remove the damage divider if there's minimal damage (the multiplier is applied on line 66, so this negates it if 1 damage is being dealt with a small number of players, such that a small party in a small encounter would still take damage instead of always getting 0 after division + rounding)

app/reducers/combat.tsx
toddmedema: Was thinking about that... any recommendations for testing code that incorporates randomness / probability? I suppose we could, at the most basic, test to make sure that the damage outputted is between the ranges of the mathematically possible (ie 0 and 2 * tier * multipliers)

smartin015: Other options:
- use a seedable RNG (https://github.com/davidbau/seedrandom), or
- split the function in two and pass in an array of "random" values into the inner function

toddmedema: Added to test issue since framework is still borked - https://github.com/Fabricate-IO/expedition-app/issues/27

PR #171, opened 2017-01-08
rate the app

toddmedema:


app/components/base/Card.tsx
```diff
@@ -36,8 +38,16 @@ export default class ExpeditionCard extends React.Component<ExpeditionCardProps,
         return getStore().dispatch(toPrevious('SPLASH_CARD', undefined, false));
       case 'SETTINGS':
         return getStore().dispatch(toCard('SETTINGS'));
+      case 'RATE':
+        if (window.platform === 'android') {
+          window.open('https://play.google.com/store/apps/details?id=io.fabricate.expedition', '_system');
```
smartin015: These could probably be constants (i.e. in constants.tsx)

app/components/base/Card.tsx
```diff
@@ -36,8 +38,16 @@ export default class ExpeditionCard extends React.Component<ExpeditionCardProps,
...
+        }
+        else if (window.platform === 'ios') {
+          window.open('https://itunes.apple.com/us/app/expedition-roleplaying-card/id1085063478?ls=1&mt=8', '_system');
+        }
+        break;
       case 'FEEDBACK':
-        var url = 'http://www.expeditiongame.com/contact/?utm_source=webapp&utm_medium=app';
+        var url = 'http://www.expeditiongame.com/contact/?utm_source=app&utm_medium=' + window.platform;
```
smartin015: Same with this URL

PR #172, opened 2017-01-08
Added docker install.sh and run.sh files

smartin015:
also agnosticized ports and updated dependencies/gitignore


toddmedema:
Need to update README, then looks good

PR #174, opened 2017-01-08
hide overflowing part of excessively long titles

toddmedema:
closes https://github.com/Fabricate-IO/expedition-app/issues/169


smartin015:
Suggest `text-overflow: ellipsis`

https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/

PR #178, opened 2017-01-15
identify encounters regardless of capitalization

toddmedema:


app/Encounters.tsx
```diff
@@ -1,243 +1,243 @@
 //GENERATED BY ./config/build_encounters.js
```
smartin015: Was this substitution done by build_encounters.js, or via the spreadsheet? If not, then this'll revert as soon as we re-run the build tool.

toddmedema: Good call, fixed

toddmedema: (updated in build_encounters - still want the names to be capitalized when displayed)

PR #179, opened 2017-01-16
add single player rules variant

toddmedema:
aka you may draw three and pick one; also streamline some of the other rules to specifically state one, since you're only one player


app/components/QuestStart.tsx
```diff
@@ -15,18 +16,21 @@ export interface QuestStartProps extends QuestStartStateProps, QuestStartDispatc
 
 // TODO: Refactor this into a QUEST_CARD
 const QuestStart = (props: QuestStartProps): JSX.Element => {
+  const singlePlayer = (props.settings.numPlayers === 1);
```
smartin015: This doesn't appear to be used... also !multiPlayer == singlePlayer.

toddmedema: @smartin015 they're both used; mainly did it to make the rules text easier to human-read

smartin015: Overlooked the usage. However, I still don't think this is really worth two separate vars. Whatevs.

PR #182, opened 2017-01-18
style for tablets

toddmedema:
Worth pulling it down on your local computer and playing with browser device emulation / different screen sizes to see how it feels :)

closes https://github.com/Fabricate-IO/expedition-app/issues/170


smartin015:
LGTM

PR #183, opened 2017-01-19
Add Firebase analytics to app

toddmedema:
Then imports the data to BigQuery, which you can query against here: https://bigquery.cloud.google.com/queries/expedition-quest-ide
(see query history -> project queries -> starter for some example queries against this data)

Note that the Firebase plugin we're using currently only supports Android and iOS, so the web app is shimmed to just log to console. Can fix that in a future release (hoping they'll just update the plugin...), but wanted to get some analytics flowing ASAP to help with game balance, understanding how much people are playing, etc

As part of the review, also open to other ideas that we could / should log :)


app/actions/web.tsx
```diff
@@ -23,8 +23,13 @@ export function loadQuestXML(data: XMLElement | string, ctx: QuestContext) {
     if (questNode.get(0).tagName.toLowerCase() !== "quest") {
       throw 'Invalid Quest - missing <quest> node';
     }
-    var firstNode = questNode.children().eq(0);
-    dispatch(initQuest(firstNode, ctx));
+
+    const firstNode = questNode.children().eq(0);
```
smartin015: Suggest moving this logic into initQuest()

PR #184, opened 2017-01-29
Feedback card at the end of quests

toddmedema:
Also changes the default port to 8081 so that you can run the app and the quest creator at the same time locally

Note that the /feedback URL doesn't currently require auth, because otherwise people couldn't send feedback on the stock quests. It also BCC's me on every feedback email sent, so we can monitor for abuse and add an auth requirement if / when it becomes necessary.


toddmedema:
Screenshots:
screen shot 2017-01-29 at 1 08 10 pm
screen shot 2017-01-29 at 1 07 59 pm

toddmedema:
@smartin015 updated to include alert on success / fail. Alerts aren't the most awesome of UI's, but it gets the job done & we can always improve later. (really want to get this out before the monthly newsletter)

README.md
```diff
@@ -58,7 +58,7 @@ To build for deploying, you'll also need `google-services.json` and `GoogleServi
 NODE_ENV=dev node ${SCRIPT:-app.js}
 ```
 
-This outputs an IP address you can use to locally test and another that can be used on devices connected to your network.
+This runs the app at `http://localhost:8081/`. It also outputs an IP address you can use to locally test and another that can be used on devices connected to your network.
```
smartin015: *Technically* this is overridden if DOCKER_PORT is defined. Probably worth a mention.

app/components/QuestEnd.tsx
```diff
@@ -0,0 +1,69 @@
...
+      <p>If you have any feedback you'd like to send to the author, now's your chance (providing feedback is option).</p>
+      <TextArea onBlur={props.onCommentChange}></TextArea>
+      <Button onTouchTap={() => props.onSubmit(props.quest, props.settings)}>
+        {(props.quest.feedback === '' || props.quest.feedback == null) ? 'Return to menu' : 'Send feedback'}
+      </Button>
+    </Card>
+  );
+}
+
+class TextArea extends React.Component<any, any> {
```
smartin015: I'd feel more comfortable if this had a prefix (e.g. `FeedbackTextArea`) since we're including specific hint text and other styling that assumes placement in the end quest card only.

app/components/QuestEnd.tsx
```diff
@@ -0,0 +1,69 @@
...
+    this.setState({value: value});
+    if (this.props.onChange) { this.props.onChange(value); }
+  }
+
+  render() {
+    return (
+      <TextField
+        fullWidth={true}
+        hintText="your feedback - what you liked, what you didn't like, bug reports, etc"
+        hintStyle={{color: '#444', top: '16px', left: '10px', right: '10px', bottom: 'auto'}}
```
smartin015: Probs better to keep styling in styles.scss and use a className attribute... IIRC we moved away from inline styling because it was very hard to trace.

app/components/QuestEndContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+    onCommentChange: (text: string) => {
+      dispatch(updateFeedback(text));
+    },
+    onSubmit: (quest: QuestState, settings: SettingsType) => {
+
+      const feedback = quest.feedback;
+      const meta = quest.details;
+
+      if (feedback.length > 0 && feedback.length < 20) {
+        return alert('More feedback required: Feedback must be at least 20 characters to provide value to the quest writer');
```
smartin015: Could we have a TODO to replace these errors with something more stylish?

Also, how do alerts show up in cordova apps? Have you tested this?

app/components/QuestEndContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+      }
+
+      if (feedback.length > 20) {
+        $.post({
+          url: authSettings.urlBase + '/feedback',
+          data: JSON.stringify({title: meta.title, authorName: meta.author, authorEmail: meta.email, feedback: feedback, players: settings.numPlayers, difficulty: settings.difficulty}),
```
smartin015: Any reason why there are some mismatches in naming here (authorName vs author, authorEmail vs email, players vs numPlayers)?

app/components/QuestEndContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+      const feedback = quest.feedback;
+      const meta = quest.details;
+
+      if (feedback.length > 0 && feedback.length < 20) {
+        return alert('More feedback required: Feedback must be at least 20 characters to provide value to the quest writer');
+      }
+
+      if (feedback.length > 20) {
+        $.post({
```
smartin015: Ideally the whole post request should be in [actions/web.tsx](https://github.com/Fabricate-IO/expedition-app/blob/master/app/actions/web.tsx) so we keep all our call/response code in the same location. See [Async Actions](http://redux.js.org/docs/advanced/AsyncActions.html)

app/components/QuestEndContainer.tsx
```diff
@@ -0,0 +1,57 @@
...
+  return {
+    onCommentChange: (text: string) => {
+      dispatch(updateFeedback(text));
+    },
+    onSubmit: (quest: QuestState, settings: SettingsType) => {
+
+      const feedback = quest.feedback;
+      const meta = quest.details;
+
+      if (feedback.length > 0 && feedback.length < 20) {
```
smartin015: 20 should be a referenced constant value (since it shows up like 3 times) and passed into the alert strings

app/components/QuestEnd.tsx
toddmedema: done

README.md
toddmedema: done

app/components/QuestEnd.tsx
toddmedema: moved what I could, but their implementation of hint text isn't specific enough to do a good job of selecting it (added a comment to that effect)

app/components/QuestEndContainer.tsx
toddmedema: good idea

app/components/QuestEndContainer.tsx
toddmedema: Added; just confirmed, alerts appear like normal browser alerts

app/components/QuestEndContainer.tsx
toddmedema: done

app/components/QuestEndContainer.tsx
toddmedema: Excellent point. Not particularly attached to them, just wrote the server / email side first and those seemed like the natural values (fwiw numPlayers seems a bit inconsistent with the rest of the naming, ie since it's the less-rigorously-specific-but-shorter `.author` and `.email`, perhaps it should be `.players`?)

README.md
```diff
@@ -58,7 +56,7 @@ To build for deploying, you'll also need `google-services.json` and `GoogleServi
 NODE_ENV=dev node ${SCRIPT:-app.js}
 ```
 
-This runs the app at `http://localhost:8081/`. It also outputs an IP address you can use to locally test and another that can be used on devices connected to your network.
+This runs the app at `http://localhost:8081/` (default port is 8081; may be different if you're using the docker container via `./install.sh` and `./run.sh`). It also outputs an IP address you can use to locally test and another that can be used on devices connected to your network.
```
smartin015: This might actually be more confusing... the install and run scripts are just vanilla linux scripts - it's really [containerizer](https://github.com/Fabricate-IO/containerizer) that does the docker port setting magic.

app/components/QuestEndContainer.tsx
smartin015: That does seem more in line with the others. Your choice to either stick with numPlayers and add a TODO, or batch replace :)

README.md
toddmedema: @smartin015 updated

PR #187, opened 2017-01-30
improve combat balance, show enemy class icon in draw list

toddmedema:
screen shot 2017-01-29 at 10 27 13 pm
screen shot 2017-01-29 at 10 26 59 pm


app/components/Combat.tsx
```diff
@@ -54,10 +54,10 @@ function renderSelectTier(props: CombatProps): JSX.Element {
 
 function renderDrawEnemies(props: CombatProps): JSX.Element {
   let enemies: JSX.Element[] = props.combat.enemies.map(function(enemy: Enemy, index: number) {
+    const icon = (enemy.class) ? `<img class="inline_icon" src="images/${enemy.class.toLowerCase()}_white_small.svg">` : '';
```
smartin015: Since we're dangerously setting HTML based on the class here, we'll need to be careful once we allow quest creators to define custom cards. I'd suggest doing a preemptive HTML filter on the enemy.class field, just in case.

app/reducers/combat.tsx
```diff
@@ -60,6 +60,7 @@ export function generateCombatAttack(combat: CombatState, elapsedMillis: number,
 
   // Scale according to multipliers, then round to nearest whole number.
   damage = Math.round(damage * combat.damageMultiplier * playerMultiplier);
+  damage = Math.min(10, damage);
```
smartin015: Might as well put this on the same line as above

app/components/Combat.tsx
toddmedema: @smartin015 thinking that'd be better off in the quest creator side / when it first renders it into XML - added as a note to the custom enemies issue: https://github.com/Fabricate-IO/expedition-quest-creator/issues/82

app/reducers/combat.tsx
toddmedema: Fixed

toddmedema:
@smartin015 updated to improve styling for long enemy names:

screen shot 2017-01-30 at 8 22 17 pm

app/components/Combat.tsx
smartin015: I don't think we should rely on correct XML escapement - if we ever discover a bug related to it then we'd have to manually disapprove quests until bad quests were republished. If we escape it here, it's much less urgent and requires less reaction if we ever discover a bug.

PR #190, opened 2017-01-31
fix touch debounce bugs

toddmedema:
closes https://github.com/Fabricate-IO/expedition-app/issues/136#issuecomment-276109418


app/reducers/card.tsx
```diff
@@ -1,11 +1,15 @@
 import {CardState} from './StateTypes'
 import {NavigateAction, ReturnAction} from '../actions/ActionTypes'
 
-export function card(state: CardState = {name: 'SPLASH_CARD', ts: 0}, action: Redux.Action): CardState {
+export function card(state: CardState = {name: 'SPLASH_CARD', ts: Date.now()}, action: Redux.Action): CardState {
   switch (action.type) {
     case 'NAVIGATE':
-      return (action as NavigateAction).to;
+      const to = (action as NavigateAction).to;
+      if (to.ts - state.ts < 400) { // prevent accidental double-carding (<400ms between inputs)
```
smartin015: This looks like it should be a const value... I'd suggest declaring it in advance of the function or before the switch statement. If descriptive enough, we wouldn't need to add the comment :)

toddmedema: Named `NAVIGATION_DEBOUNCE_MS`

PR #193, opened 2017-02-02
bugifixing editor behavior

toddmedema:
@smartin015 looks like [adding Firebase to the app](https://github.com/Fabricate-IO/expedition-app/pull/183/files) broke how it interfaces with the Quest Creator.

When I try to use play from cursor, I get the error:
```
Uncaught TypeError: Cannot read property 'details' of undefined
    at Object.n [as quest] (http://devquests.expeditiongame.com/bundle.js:57:3817)
    at r (http://devquests.expeditiongame.com/bundle.js:21:3070)
    at Object.o [as default] (http://devquests.expeditiongame.com/bundle.js:21:3775)
    at r (http://devquests.expeditiongame.com/bundle.js:48:1121)
    at computeNextEntry (<anonymous>:2:25105)
    at recomputeStates (<anonymous>:2:25405)
    at <anonymous>:2:28958
    at Object.c [as dispatch] (http://devquests.expeditiongame.com/bundle.js:54:16613)
    at dispatch (<anonymous>:2:29451)
    at http://devquests.expeditiongame.com/bundle.js:54:15538
    at Object.s [as loadNode] (http://devquests.expeditiongame.com/bundle.js:7:7052)
    at Object.playFromCursor (http://devquests.expeditiongame.com/bundle.js:55:2615)
    at Object.onTouchTap (http://devquests.expeditiongame.com/bundle.js:55:1952)
```

and when I try to use play from start, I get the error:
```
Uncaught TypeError: r.getAttribute is not a function
    at Object.i [as initQuest] (http://devquests.expeditiongame.com/bundle.js:7:5921)
    at Object.playFromStart (http://devquests.expeditiongame.com/bundle.js:55:2771)
    at Object.onTouchTap (http://devquests.expeditiongame.com/bundle.js:55:2071)
```

Which seems to be related to how the quest creator is sending the calculated quest, vs the app pulling it from the database.

Should probably figure out how to fix on the quest creator side since that's easier to deploy updates to


toddmedema:
Specifically, seems to be something about how it changed the expectations around the incoming quest node https://github.com/Fabricate-IO/expedition-app/pull/183/files#diff-71b1ccd0346da110fc5d3d6f6df69cf8R26

Oddly, even when I fix the issue (and, perhaps I'm fixing it incorrectly), hitting play from start doesn't throw any errors...but also doesn't do anything

toddmedema:
Note from IRL: likely due to the quest creator passing a cheerio object instead of a jquery object, which is a strict subset of jquery. So, we can update the app to use the cheerio subset - such as using .attr instead of .getAttribute - http://api.jquery.com/attr/

app/actions/quest.tsx
```diff
@@ -21,21 +21,23 @@ export function handleCombatTimerStop(elapsedMillis: number, settings: SettingsT
...
+    author: metaNode.attr('author'),
+    email: metaNode.attr('email'),
+    ur: metaNode.attr('url'),
+    minPlayers: metaNode.attr('minPlayers'),
+    maxPlayers: metaNode.attr('maxPlayers'),
+    minTimeMinutes: metaNode.attr('minTimeMinutes'),
+    maxTimeMinutes: metaNode.attr('maxTimeMinutes'),
   };
-  window.FirebasePlugin.logEvent("quest_start", meta);
+  if (!editor) {
```
smartin015: I think it would be nicer to move the quest_start logging into the quest loading code (which is not used by the IDE) so we don't have to pass an editor flag. We'll probably want to change the event name to "quest_load" or some such though. Note that this can be done by calling initQuest() and logging its details attribute before dispatching.

app/reducers/card.tsx
```diff
@@ -2,7 +2,7 @@ import {CardState} from './StateTypes'
 import {NavigateAction, ReturnAction} from '../actions/ActionTypes'
 import {NAVIGATION_DEBOUNCE_MS} from '../constants'
 
-export function card(state: CardState = {name: 'SPLASH_CARD', ts: Date.now()}, action: Redux.Action): CardState {
+export function card(state: CardState = {name: 'SPLASH_CARD', ts: 0}, action: Redux.Action): CardState {
```
smartin015: Since this was an obtuse bug, maybe add a comment indicating that ts=0 because of quest IDE?

app/reducers/quest.tsx
```diff
@@ -2,6 +2,7 @@ import {QuestState, AppState} from './StateTypes'
 import {QuestNodeAction, InitCombatAction, UpdateFeedbackAction} from '../actions/ActionTypes'
 
 export function quest(state: QuestState, action: Redux.Action): QuestState {
+  state = state || {};
```
smartin015: Did this actually fix anything? If so, better to use the default parameter assignment as in card.tsx - if not, let's remove it.

app/actions/quest.tsx
toddmedema: Good idea, updated here and in the QC PR

app/reducers/card.tsx
toddmedema: updated

app/reducers/quest.tsx
toddmedema: Might've "fixed" it when we'd gotten 1 out of 2 bugs, but looks like it's unnecessary now, good catch

PR #203, opened 2017-02-04
improved touch trigger

toddmedema:
closes https://github.com/Fabricate-IO/expedition-app/issues/197

Changes the touches to be (almost) black, with a more prominent white circle around them. Fits better with the theme. Note that the combat timer page is black, but it still seemed to look fine with the white circle (you can also try it out on your device at beta.expeditiongame.com)

Speeds up the drawing using animation frames. No longer flickering, though redraws are still slow / laggy on really old devices - but it does look much better without the flickering!


app/components/base/MultiTouchTrigger.tsx
```diff
@@ -5,10 +5,16 @@ interface MultiTouchTriggerProps extends React.Props<any> {
   onTouchChange: (touches: any) => any;
 }
 
+const CENTER_RADIUS = 36;
```
smartin015: Thanks for pulling these out into constants. Can we put them all in a JSON const object (e.g. traditional react `const style = {...}` format)?

PR #204, opened 2017-02-04
improved helptext

toddmedema:
Got feedback from a user re: learning to adventure quest: "Damage buffs could be explained more thoroughly"

So, don't know if this is what they were talking about exactly, but went through and cleaned up some of the tutorial text


app/quests/learning_to_adventure.xml
```diff
@@ -69,10 +69,13 @@
   </roleplay>
   <roleplay title="A Magical Fight" id="magicfight">
     <instruction>
-      <p>You're about to enter combat for the first time.</p>
+      <p>[fae] You're about to enter combat for the first time.</p>
     </instruction>
     <instruction>
-      <p>The app will guide you through each round. Just remember: you must reduce the encounter tier to 0 (defeat all enemies) to win. If all adventurers are defeated, you lose.</p>
+      <p>The app will guide you through each round. To win, you must reduce the encounter tier to 0 (defeat all enemies). If all adventurers are knocked out, you lose.</p>
```
smartin015: "you lose" - should probably indicate that sometimes this doesn't mean you lose the quest.

toddmedema: updated

PR #211, opened 2017-02-07
Improved Feedback

toddmedema:
- include user email and platform in feedback message
- with user email, ask them (opt-in) if they'd like to share their email with the quest creator
- if they opt in to sharing, but aren't logged in yet, prompt to log in

pairs with https://github.com/Fabricate-IO/expedition-quest-creator/pull/176


toddmedema:
NOTE: this incorporates the onlinefow PR, which is why it looks so big

PR #217, opened 2017-02-12
add a bit more choice to learning to play quest

toddmedema:


smartin015:
Could I get a link to the quest doc?

toddmedema:
http://quests.expeditiongame.com/#0B7K9abSH1xEOeEZSVVMwNHNqaFE

(note that all featured quests are in Drive/Fabricate/Expedition/quests)

PR #228, opened 2017-03-01
add logarithmic curve to loot

toddmedema:
closes https://github.com/Fabricate-IO/expedition-app/issues/227


toddmedema:
@smartin015 hmmm, looks like all tests are currently commented out? Did that happen from the TypeScript upgrade? / would you be willing to create a working example test case? If so, I can take it from there + re-implement existing test cases

PR #231, opened 2017-03-01
Condense victory screen to fit without scrolling (on most devices)

toddmedema:
Closes https://github.com/Fabricate-IO/expedition-app/issues/224

screen shot 2017-02-28 at 7 56 05 pm


toddmedema:
@smartin015 updated

app/constants.tsx
```diff
@@ -27,3 +27,29 @@ export const NAVIGATION_DEBOUNCE_MS = 600;
...
+    case 1: return 'one';
+    case 2: return 'two';
+    case 3: return 'three';
+    case 4: return 'four';
+    case 5: return 'five';
+    case 6: return 'six';
+    case 7: return 'seven';
+    case 8: return 'eight';
+    case 9: return 'nine';
+    case 10: return 'ten';
```
smartin015: I'd add a default here that just returns the number without english translation. Otherwise, you'll probably fail out and not display anything at all.

toddmedema: good call - default cause also removes the need for the three lines at the top!

PR #232, opened 2017-03-01
add support for _. random encounter functions

toddmedema:
Such as `<e>{{_.randomEnemyOfClassTier("bandit", 1)}}</e>`

Note: might be nice to hook them up to `_.encounters` value, so that writer modifications to that list would be reflected.

Note: mathjs only works with double quotes for strings, throws weird errors if you try to use single quotes

Closes https://github.com/Fabricate-IO/expedition-quest-creator/issues/196


app/constants.tsx
```diff
@@ -51,3 +51,9 @@ export function numberToWord(input: number): string {
 export function capitalizeFirstLetter(input: string): string {
   return input.charAt(0).toUpperCase() + input.slice(1);
 }
+
+// return a random property from the provided object
+export function randomProperty(obj: any): any {
+  const keys = Object.keys(obj);
+  return obj[keys[ keys.length * Math.random() << 0]];
```
smartin015: I'd much prefer `Math.floor` to left-shiftery. It wasn't immediately obvious the shift here was a cast.

app/constants.tsx
```diff
@@ -51,3 +51,9 @@ export function numberToWord(input: number): string {
 export function capitalizeFirstLetter(input: string): string {
   return input.charAt(0).toUpperCase() + input.slice(1);
 }
+
+// return a random property from the provided object
+export function randomProperty(obj: any): any {
```
smartin015: Suggest `randomPropertyValue` instead of `randomProperty`, as it's not clear you're referring to the key or value of the prop

app/reducers/QuestTypes.tsx
```diff
@@ -32,7 +35,30 @@ export interface QuestContext {
...
+  return {scope: {
+    _: {
+      encounters: encounters,
+      randomEnemy: function(): string {
+        return randomProperty(encounters).name;
+      },
+      randomEnemyOfTier: function(tier: number): string {
+        const encountersNames = Object.keys(encounters)
+          .filter( key => encounters[key].tier === tier );
+        return encountersNames[Math.floor(Math.random() * encountersNames.length)];
```
smartin015: Why not just `randomProperty(encountersNames)`? For that matter, randomProperty could also take a lambda that filters the keys of the object as well. You'd need to use `randomProperty(encounters, lambdafn).name.toLowerCase()` to get the same result, though.

app/constants.tsx
toddmedema: Done

app/reducers/QuestTypes.tsx
toddmedema: done

app/constants.tsx
toddmedema: done

PR #241, opened 2017-03-07
add _.numAdventurers()

toddmedema:


app/reducers/QuestTypes.tsx
```diff
@@ -39,6 +40,9 @@ export function defaultQuestContext(): QuestContext {
   // added carefully b/c we'll have to support them going forward.
   return {scope: {
     _: {
+      numAdventurers: function(): number {
+        return getStore().getState().settings.numPlayers;
```
smartin015: This could benefit from a quick check that `settings` is defined. Ideally it's always defined, but as we're presenting a layer to the user and Redux is hard, we shouldn't bank on that.

app/reducers/QuestTypes.tsx
```diff
@@ -1,5 +1,6 @@
 import {randomPropertyValue} from '../constants'
 import {encounters} from '../Encounters'
+import {getStore} from '../store'
```
smartin015: It makes me sad to punch yet another hole through Redux abstraction :( up until now, the only times we've reached in to the store directly was at top-level components like Main and Card.

Couldn't we instead apply state via reducer? The context is already updated via CombatResult/RoleplayResult...

app/reducers/QuestTypes.tsx
toddmedema: done

app/reducers/QuestTypes.tsx
toddmedema: @smartin015 hmm. Not sure if reducer is the right tool? We aren't modifying the state, just reading it (otherwise I'd agree that something fancier might be better)

smartin015: I have some ideas about this that I'll put in a bug. For now, we can go with what's easy.

PR #242, opened 2017-03-08
add dungeon crawl to advanced modes

toddmedema:
includes fix for Quest Creator using JSON attributes on events


toddmedema:
Updated home screen:
screen shot 2017-03-07 at 10 34 21 pm

app/QuestParser.tsx
```diff
@@ -60,34 +61,40 @@ export function init(root: XMLElement): XMLElement {
...
-    let value = child.attributes[i].value;
-    switch (key) {
-      case 'xp': value = (value == 'true'); break;
-      case 'loot': value = (value == 'true'); break;
-      case 'heal': value = parseInt(value); break;
+  const node = getEvent(parent, event, ctx).get(0);
+  const cheeriod = Cheerio(node).get(0);
+  if (cheeriod) {
+    return _cleanParams(cheeriod.attribs);
+  } else {
```
smartin015: In what case does cheerio not return anything useful?

app/reducers/StateTypes.tsx
```diff
@@ -4,25 +4,25 @@ import {CombatResult, RoleplayResult} from '../QuestParser'
...
 
 export interface DOMElement {
-  attributes: any; //NamedNodeMap
+  attributes: any; // NamedNodeMap
   outerHTML: string;
   parentNode: DOMElement;
   tagName: string;
 };
 
 export interface XMLElement {
```
smartin015: I suspect this is also used in the quest creator... I'd be careful about getting rid of elements here.

toddmedema: (not eliminating, just added a space to the comment)

app/QuestParser.tsx
toddmedema: When running the app as just the app - added comments

app/reducers/StateTypes.tsx
smartin015: Didn't grok that this was a rearrangement... whoops :)

app/QuestParser.tsx
```diff
@@ -60,34 +61,42 @@ export function init(root: XMLElement): XMLElement {
...
-      case 'xp': value = (value == 'true'); break;
-      case 'loot': value = (value == 'true'); break;
-      case 'heal': value = parseInt(value); break;
+  const node = getEvent(parent, event, ctx).get(0);
+  const cheeriod = Cheerio(node).get(0);
+  // from Quest Creator
+  if (cheeriod) {
+    return _cleanParams(cheeriod.attribs);
+  // from App
+  } else {
```
smartin015: As we talked about, this is really a hack since we don't know why the IDE is providing differently-formatted XML than the app itself. Plz add a TODO indicating so and link to a bug to track the issue.

app/QuestParser.tsx
```diff
@@ -60,34 +61,42 @@ export function init(root: XMLElement): XMLElement {
...
+    const params: any = {};
+    const attributes = node.attributes;
+    for (let i = 0; i < attributes.length; i++) {
+      params[attributes[i].name] = attributes[i].value;
     }
-    params[key] = value;
+    return _cleanParams(params);
+  }
+
+  function _cleanParams(obj: any) {
```
smartin015: It's not clear whether assigning to the XMLElement's attributes actually changes the element or not... can you do a new object definition, e.g:

```
return {
  "xp": ...
...
}
```

And also add a return type, so we don't break typescript for this and everything downstream.

Issue #230, opened 2017-03-01
Make text non-selectable


toddmedema:
Super annoying to have to fight against the text selection interface when using the app

PR #247, opened 2017-03-10
disable text selection

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/230

Needs to be tested on a working copy of the repo to confirm, but hand test in console confirmed the CSS is correct


toddmedema:
Confirmed, doesn't break text inputs

Issue #160, opened 2016-12-23
Text input


smartin015:
*Migrated to expeditionrpg/expedition#186 by [spacetelescope/github-issues-import](https://github.com/spacetelescope/github-issues-import)*

Now that we have ops, we can also operate on more sophisticated user input than single button clicks.

- text input (secret passphrases, or fine value manipulation)
- multi-selection (perhaps for a better flow for purchasing items)
- sliders
- canvas-based input

Text input especially would allow more personalized stories, and thus more character building.

toddmedema:
Seems like text input or `select` is a good starting point

smartin015:
Backlogging for now

PR #262, opened 2017-03-15
make roleplay cards more flexible; add support for instructions anywhere in card

toddmedema:
Instructions can now be displayed anywhere, not just grouped at the bottom.

Also lays the foundation for https://github.com/ExpeditionRPG/expedition-app/issues/160 by allowing more flexible types of elements in a roleplaying card

screen shot 2017-03-15 at 5 04 49 pm


app/QuestParser.test.tsx
```diff
@@ -12,24 +12,25 @@ describe('QuestParser', () => {
...
       // Lines with nothing but variable assignment are hidden
       var result = loadRoleplayNode(cheerio.load('<roleplay><p>{{text="TEST"}}</p><p>{{text}}</p></roleplay>')('roleplay'), {scope: {}});
-      expect(mount(result.content).html()).toEqual('<span><p>TEST</p></span>');
+      expect(result.content[0].text).toEqual('<p>TEST</p>');
+      expect(result.content.length).toEqual(1);
 
       // variables with value 0 and 1 display properly
       var result = loadRoleplayNode(cheerio.load('<roleplay><p>{{a=0}}{{b=1}}</p><p>{{a}}{{b}}</p></roleplay>')('roleplay'), {scope: {}});
-      expect(mount(result.content).html()).toEqual('<span><p>01</p></span>');
+      expect(result.content[0].text).toEqual('<p>01</p>');
```
smartin015: I'd prefer something like:

```
expect(result.content).toEqual([{text: '<p>01</p>'}]);
```

Here and elsewhere. That way we're testing on the full body of text and not ignoring if anything somehow happens after.

app/QuestParser.tsx
```diff
@@ -392,16 +391,12 @@ function generateIconElements(content: string): string {
...
+  let numEvents = 0;
+  let choices: Choice[] = [];
+  let choiceCount = -1;
+  let children: RoleplayElement[] = [];
 
-  var newScope = Object.assign({}, ctx.scope);
-
-  var choiceCount = -1;
-  var instructionCount = -1;
+  const newScope = Object.assign({}, ctx.scope);
```
smartin015: Spread? :D

app/QuestParser.tsx
```diff
@@ -427,33 +422,41 @@ export function loadRoleplayNode(node: XMLElement, ctx: QuestContext): RoleplayR
...
 
-    // accumulate 'instruction' tags in tags[]
+    let type = 'text';
+    let textContent = '';
     if (tag === 'instruction') {
-      instructionCount++;
-      instructions.push({text: generateIconElements(evaluateContentOps(c.html(), newScope)), idx: instructionCount});
-      return;
+      type = 'instruction';
+      textContent = c.html();
```
smartin015: Random thought... it looks like this allows users to pass through arbitrary HTML code. Are we doing any app-side sanitizing?

app/QuestParser.tsx
```diff
@@ -427,33 +422,41 @@ export function loadRoleplayNode(node: XMLElement, ctx: QuestContext): RoleplayR
...
-      textContent = c.toString();
-    } else {
-      throw new Error('Invalid element ' + c);
+    textContent = evaluateContentOps(textContent, newScope);
+
+    if (textContent !== '') {
+      children.push({
+        type,
+        text: generateIconElements(textContent),
+        idx: children.length,
```
smartin015: Why are we enumerating objects in a list?

app/reducers/QuestTypes.tsx
```diff
@@ -85,7 +85,8 @@ export interface EventParameters {
   heal?: number;
 }
 
-export interface Instruction {
+export interface RoleplayElement {
+  type: string; // text, instruction
```
smartin015: Instead: `type: 'text' | 'instruction';`

app/QuestParser.tsx
toddmedema: Hmm. Should we be doing that in the app or quest creator? Pretty sure we're already doing that in the QC

app/reducers/QuestTypes.tsx
toddmedema: Wait, really? That's awesome

app/QuestParser.tsx
toddmedema: Touche

app/QuestParser.tsx
toddmedema: 👍

app/QuestParser.test.tsx
toddmedema: @smartin015 updated this; all comments addressed

app/QuestParser.tsx
smartin015: I'd say both. We don't know what input sources we'll end up using aside from checked-in quests, and there might be ways of adding quests later on that don't go through the sanitization (for instance, direct calls to /publish without going through the editor). 

But it also benefits us to not have latent unsanitized html code in our DB waiting for an exploit.

app/QuestParser.tsx
smartin015: Doesn't look like this was spreaded...

toddmedema:
Fixed the last spreads; mergin'

PR #267, opened 2017-03-19
add scope._.views.<id>

toddmedema:
Main questions:
- is .views the best name? (vs, say, .viewCount)
- is this the best way to implement it? Say, vs a wrapping function that returns 0 if undefined
  - perhaps _.views gets updated with the raw number, then _.viewCount('id') would return the number, with a default of 0?


toddmedema:
@smartin015 further testing on context issue has revealed some interesting things:

1. happens on master / before this PR
2. state.quest.result.ctx.scope doesn't seem to get updated on RETURN / backs, even though the info displayed in the quest card (ie {{foo}}) usually displays the old / updated value on backs.
3. state.quest.result.ctx.views seems to be a bit less consistent here, if you go back to the start of the quest, it's consistently 0, but if you only go back a card or two, it keeps incrementing

...but, the place in the code that does the RETURN is generic / shouldn't result in different experiences for different parts of the state: https://github.com/ExpeditionRPG/expedition-app/blob/master/app/reducers/CombinedReducers.tsx#L37

Any ideas?

toddmedema:
Alright, so the scope part was just me being dumb, you have to click on the new entry in the state history and then the scope is correct / as expected #DevTools

But.....views is behaving incorrectly / not resetting with going back. Any ideas @smartin015 ?

smartin015:
As we discussed... I think we're going to need a deep copy instead of just using spread to copy the object. Otherwise we end up with a bunch of references.

toddmedema:
@smartin015 yep, that did it! everything else look good?

app/reducers/QuestTypes.tsx
```diff
@@ -33,40 +33,48 @@ export type Loot = {tier: number, count: number};
 export interface QuestContext {
   // Scope is passed to the parser when rendering
   // nodes that are potentially parseable via MathJS.
-  scope: any; //TODO: required fields later
+  scope: any; // TODO: required fields later
+  views: any; // TODO: {string: number}
```
smartin015: Why not just type them now?

PR #275, opened 2017-03-24
update manifest, webpack

toddmedema:
Non-breaking, merging now for later review

Update / expand manifest.json
Have webpack properly copy it for deploys
(also, fix a quick bug with the noscript tag, sorry)


app/react.tsx
```diff
@@ -96,6 +96,7 @@ if (window.FirebasePlugin) {
 
 // DOM ready
 $(() => {
+  $('#noscript').hide();
```
smartin015: `<noscipt>` => `<noscript>`, and remove

PR #277, opened 2017-03-25
shuffle all abilities, except last round

toddmedema:
We talked about this at UnPub - wanted to put it in PR form to continue the conversation


smartin015:
I like it - but let's playtest first.

toddmedema:
@smartin015 updated based on playtest

PR #280, opened 2017-03-27
fix bug with loot max tier carrying over between encounters

toddmedema:
turns out the definition of combat changed (no longer nulled between fights)
can either do it this way using phase detection, or we can figure out why it's no longer getting nulled (If we wanted nulling to be the case)


app/reducers/combat.tsx
```diff
@@ -149,21 +148,18 @@ export function combat(state: CombatState, action: Redux.Action): CombatState {
...
         levelUp: (victoryAction.numPlayers <= victoryAction.maxTier),
+        loot: generateLoot(victoryAction.maxTier),
       };
     case 'TIER_SUM_DELTA':
       let newTierCount = state.tier + (action as TierSumDeltaAction).delta;
-      if (newTierCount < 0) {
-        return state;
-      }
-      return {...state, tier: newTierCount};
+      return {...state, tier: Math.max(newTierCount, 0)};
```
smartin015: Good call. I could see newTierCount getting pushed below zero for some reason, and the user getting stuck with the old code.

toddmedema:
@smartin015 not sure - the app testing stack has been broken for a while. Added a stub, though, so we'll remember when we fix it

Issue #17, opened 2016-04-26
Add "Roll for party" batch random number generator to resolve screen


toddmedema:
Should be consistent even if you go back (for reference): generate on timer end.

wls:
I had toyed with the idea of making it so when you touch the d20 icon, it rolls one.  But if it's going to be part of the resolve screen, even better.

toddmedema:
Continue to hear people suggest this

Also, heard a cool suggestion - on top of a button to roll, you could also "shake to roll"

PR #281, opened 2017-03-28
add auto-roll setting

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/17

Screenshots:

screen shot 2017-03-27 at 10 47 41 pm
screen shot 2017-03-27 at 10 48 09 pm
screen shot 2017-03-27 at 10 41 07 pm


app/components/CombatContainer.tsx
```diff
@@ -65,12 +65,12 @@ const mapDispatchToProps = (dispatch: Redux.Dispatch<any>, ownProps: any): Comba
       dispatch(combatDefeat());
     },
     onTimerStop: (cardName: CardName, elapsedMillis: number, settings: SettingsType, surge: boolean) => {
+      dispatch(handleCombatTimerStop(elapsedMillis, settings));
```
smartin015: Why was this moved?

app/reducers/combat.tsx
```diff
@@ -95,14 +95,23 @@ export function generateLoot(maxTier: number): Loot[] {
...
   return loot;
-};
+}
+
+function generateRolls(count: number): number[] {
+  const rolls = [];
+  for (let i = 0; i < count; i++) {
+    rolls.push(Math.floor(Math.random() * 20) + 1);
+  }
+console.log(count, rolls);
```
smartin015: Extraneous console log

toddmedema: fixed

app/components/CombatContainer.tsx
smartin015: Something to keep in mind is that toCard also checkpoints the history... if we weren't skipping timer cards, this would cause skew of e.g. roundCount (gets incremented, history checkpoint. User clicks back button to timer card, stops timer again and increments roundCount). 

I *think* this is OK since we're skipping over the timer card on back action, but I'd recommend double-checking to be sure.

PR #283, opened 2017-03-29
fix iOS max call stack bug

toddmedema:
Tested on iOS and fixes the bug; submitted to Apple since they take a while to approve.

Turned out to be a bug with the stars that we were importing:
`import {ToggleStarBorder, ToggleStar} from 'material-ui/svg-icons'`

Which was copied from the library I was imitating

Switched to 
```
import Star from 'material-ui/svg-icons/toggle/star'
import StarBorder from 'material-ui/svg-icons/toggle/star-border'
```

And the error went away. Of course, only figured this out after upgrading StarRating to the better / more modern class structure.... hence the larger-than-minimally-necessary PR (apologies).


smartin015:
LGTM

PR #287, opened 2017-03-31
Misc small fixes

toddmedema:
- mathjs is also being required as a node module, a better way of importing it
- removed unused package imports, and removed some special webpack code for those modules (from back when I was importing star ratings)
- set process env to dev on dev, and move process env setting to the start of the webpack pipeline


webpack.config.js
```diff
@@ -42,12 +42,15 @@ const options = {
...
       { test: /\.tsx$/, loaders: ['react-hot-loader/webpack', 'awesome-typescript-loader'], exclude: /node_modules/ },
-      { test: /\.js$/, loaders: ['react-hot-loader/webpack', 'awesome-typescript-loader'], include: /node_modules\/react-native-star-rating/ },
       { enforce: 'post', test: /\.tsx$/, loaders: ['babel-loader'], exclude: /node_modules/ },
-      { enforce: 'post', test: /\.js$/, loaders: ['babel-loader'], include: /node_modules\/react-native-star-rating/ },
     ],
   },
   plugins: [
+    new webpack.DefinePlugin({
+      'process.env': {
+        NODE_ENV: JSON.stringify('dev'),
```
smartin015: It looks like this overrides **all** environment variables set (e.g. docker ports). Is that not the case? I'd also suggest casing this on whether or not there's already a NODE_ENV environment var set, so we can test production codepaths locally.

toddmedema: Good point. Removed that from this PR since this is meant to be small cleanups / non-functional changes, but did come across some webpack documentation with some good ideas about how to handle different env situations: https://webpack.js.org/guides/production-build/

Issue #288, opened 2017-04-01
Prepare parser code for IDE runtime checker


smartin015:
We should standardize how we interact with quest XML and reduce the amount of interface bloat, so we can more reliably walk the quest as if we were the user... and break less stuff when we change how quests are parsed over. 

See https://github.com/ExpeditionRPG/expedition-quest-creator/issues/68 for IDE-side bug.

smartin015:
Marking as complete. Woot!

PR #289, opened 2017-04-01
Intermediate cleanup

smartin015:
#288

This is a step in the direction of full quest runtime checking. which also highlights how end-to-endy our unit tests for the parser currently are.

No business logic changes in this PR - only exporting of functions and changing of imports (to keep diff awfulness down)


toddmedema:
LGTM

PR #292, opened 2017-04-01
Convert parser node methods into a single class

smartin015:
#288

`ParserNode` will soon be the kernel of state that's passed about by handlers etc. Eventually choice and event logic will end up in there as well, with construction of `CombatResult`, `RoleplayResult` etc. still being done in `Handlers.tsx`

Also added a todo for using the validation code in the app.


toddmedema:
Tests are breaking, but once that's fixed, LGTM

PR #295, opened 2017-04-02
Push op rendering into ParserNode

smartin015:
This is a dependent PR on #294. Changes from de411e7 onward are new.

This effectively hides *all* op string parsing/rendering for cards behind ParserNode. This means that any `loadNode` style code (e.g. `loadCombatNode`) no longer needs to worry about contextual ops.

### DIFF
--- a/app/parser/Context.tsx
+++ b/app/parser/Context.tsx
@@ -6,13 +6,16 @@ const Clone = require('clone');
 const HtmlDecode = (require('he') as any).decode;
 const Math = require('mathjs') as any;
 
-// Run MathJS over all detected {{operations}}
+// Run MathJS over all detected {{operations}}.
 export function evaluateContentOps(content: string, ctx: QuestContext): string {
   // {{.+?(?=}})}}       Match "{{asdf\n1234}}"
   // |                   Or
   // .+?(?={{|$)         Nongreedy characters (including whitespace) until "{{" or end of string
   // /g                  Multiple times
   const matches = content.match(/{{[\s\S]+?(?=}})}}|[\s\S]+?(?={{|$)/g);
+  if (!matches) {
+    return content;
+  }
 
   let result = '';
   for (let m of matches) {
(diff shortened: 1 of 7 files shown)


toddmedema:
LGTM. As noted on the other one, we'll want to test the final state of this mega-change in the quest creator at some point :)

app/parser/Context.tsx
```diff
@@ -6,13 +6,16 @@ const Clone = require('clone');
...
 const Math = require('mathjs') as any;
 
-// Run MathJS over all detected {{operations}}
+// Run MathJS over all detected {{operations}}.
 export function evaluateContentOps(content: string, ctx: QuestContext): string {
   // {{.+?(?=}})}}       Match "{{asdf\n1234}}"
   // |                   Or
   // .+?(?={{|$)         Nongreedy characters (including whitespace) until "{{" or end of string
   // /g                  Multiple times
   const matches = content.match(/{{[\s\S]+?(?=}})}}|[\s\S]+?(?={{|$)/g);
```
toddmedema: @smartin015 fun fact, apparently when Travis deploys this to prod, something breaks - but it doesn't break when rendered locally. Specifically, got this error: `Uncaught TypeError: Cannot read property 'match' of undefined`

I manually rendered and deployed to prod for now so that the web app isn't broken, but it'd be good for us to figure out what's going on.... otherwise it'll keep breaking on every deploy :/

Could potentially check to see if `content == null, return '';` before calling `content.match`, but I don't know if that's the best solution / why content would be null at all...

smartin015: I suspect it's something to do with getting XML attributes again... 
https://github.com/ExpeditionRPG/expedition-app/blob/ac83649d0d66a3ed9bc2c61318efebe900aac597/app/parser/Node.tsx#L84

Can you try checking if c.attr(j) is ever undefined here in travis-land?

toddmedema: Seems to me like we can "close" this now that deploying no longer depends on Travis, and (per your comment), your refactoring likely fixes this anyways

smartin015: Agreed.

PR #301, opened 2017-04-05
add beta.sh and prod.sh deploy scripts

toddmedema:
Tested deploys to beta and prod, both working!

Note that once this is merged into master, we'll want to run both ./beta and ./prod just in case travis snuck in a final deploy


toddmedema:
@smartin015 that was apparently a temporary glitch, re-ran the build and it worked fine

PR #302, opened 2017-04-05
Plumbed usage of ParserNode through to Roleplay.tsx

smartin015:
This keeps as much of the combat side of things the same, while using ParserNode in roleplay nodes. 

Also appearing in this pr is ParserNode.getTag(), a shortcut that makes sure we're comparing the same thing every time (no uppercase vs lowercase issues, for instance)

I also moved the rendering section of loadRoleplayNode to within Roleplay.tsx, as UI stuff shouldn't really be a part of the parser. XML attribute parsing is also broken out now and should present a cheerio/jquery agnostic view of things.


toddmedema:
LGTM

PR #304, opened 2017-04-09
Genericize combat to be a card template

smartin015:
This removes the entire `combat` reducer and custom actions, in favor of handling combat logic as updates to the quest context. Future card templates will behave similarly (tracking some state in the `QuestContext` but no dedicated reducer/actions). Excluding tests, we have about 100 fewer lines of code as a result.

I moved the `combat` actions and business logic out into a seperate file. Eventually, we should have a folder per template containing the actions, element, container, and types (basically everything that the template implements) so that we can be more modular about adding/removing new templates.

Custom combat is now practically a mini-quest, and custom combat logic is almost entirely removed (with the exception of `initCustomCombat`).

This also demonstrates testing of redux actions (both synchronous and async). We should apply this testing to the other actions as well.


smartin015:
Yes, but I wanted to limit changes during this PR. That's the next step :)

toddmedema:
Makes sense :-) lgtm!

PR #306, opened 2017-04-10
Move roleplay card into cardtemplates, abstract template usage

smartin015:
This should be the last step to almost fully partitioning actual card action handling from other non-quest components and state.


smartin015:
Closes #288

PR #313, opened 2017-04-15
add help text explanation for cases with multiple of a single enemy

toddmedema:


app/cardtemplates/combat/Combat.tsx
```diff
@@ -55,8 +55,15 @@ function renderSelectTier(props: CombatProps): JSX.Element {
 }
 
 function renderDrawEnemies(props: CombatProps): JSX.Element {
-  let enemies: JSX.Element[] = props.enemies.map(function(enemy: Enemy, index: number) {
+  let enemyNames: string[] = [];
```
smartin015: This would be a bit cleaner as a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)

In typescript, the type assignment is something like `let enemyNames: Set<string> ...`

PR #315, opened 2017-04-15
Add a few tests

smartin015:
These can be examples for other testing of actions, components, component containers, and reducers.


app/reducers/Card.test.tsx
```diff
@@ -1,4 +1,59 @@
...
+    var fixedNow = Date.now();
+    spyOn(Date, 'now').and.callFake(function() {
+      return fixedNow;
+    });
+
+    var state = card(undefined, toCard('SEARCH_CARD'));
+
+    fixedNow += 500 // ms
+    expect(card(state, toCard('QUEST_CARD'))).toEqual(jasmine.objectContaining({
+      name: 'SEARCH_CARD',
```
toddmedema: Shouldn't this be QUEST_CARD since you're going toCard QUEST_CARD?

smartin015: Good point - turns out NAVIGATION_DEBOUNCE_MS was at 600ms :) should be fixed now.

PR #319, opened 2017-04-22
End -> The End

toddmedema:
Proposed change, changing generic "End" button to the more story-like "The End"

Example:

screen shot 2017-04-22 at 3 03 23 pm


smartin015:
LGTM... after you fix tests :)

Issue #233, opened 2017-03-01
Sign in -> option to join the monthly Expedition mailing list


toddmedema:
- add checkmark to app UI (opt-in) <- only if user.email exists, doesn't always seem to exists....
  - Join the mailing list - only one email per month, highlighting new quests and features!
- post-login, once we have their email (if it's non-null), send an (authed!) "/subscribe" POST to API server
  - that adds their email to the (main) mailchimp list, not the creators list

PR #324, opened 2017-04-29
add subscribe prompt

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/233

screen shot 2017-04-29 at 4 33 04 pm


app/components/SearchContainer.tsx
```diff
@@ -28,6 +38,9 @@ const mapDispatchToProps = (dispatch: Redux.Dispatch<any>, ownProps: any): Searc
     onSearch: (numPlayers: number, user: UserState, request: SearchSettings) => {
       dispatch(search(numPlayers, user, request));
     },
+    onSubscribeChange: (v: boolean) => {
+      dispatch(changeSettings({subscribe: v}));
```
smartin015: It looks like there's no reducer code or mods to the `changeSettings` dispatch method to handle this setting change.

toddmedema: That's because the reducer is implemented generically, you just pass the dict containing the change and it applies it to the settings store

PR #325, opened 2017-04-30
improve search page styles

toddmedema:
Add icons to improve visual differentiation of time + player count
Show quest summary instead of author name, since that's much more useful for determining if you want to play

screen shot 2017-04-30 at 6 27 38 pm


toddmedema:
Updated: streamlined time units, capped descriptions to 140 character / tweet length, added tests

screen shot 2017-04-30 at 9 01 06 pm

Issue #320, opened 2017-04-23
fix iOS XML parsing - ideally, switch app to Cheerio


toddmedema:
In the latest branch, can't load quests on iOS:
```
TypeError: e.get(0).attributes[Symbol.iterator] is not a function. (In 'e.get(0).attributes[Symbol.iterator]()', 'e.get(0).attributes[Symbol.iterator]' is undefined)
```
Which is to say, `e.get(0).attributes)` in getNodeAttributes in Node.tsx

Seems to be because the iOS XML object has a different API than the web / Android AND Cheerio.

Ideally, this would be magically solved by switching the app over to use Cheerio solely and entirely

PR #326, opened 2017-05-04
fix iOS XML and subscribe bugs, search filter header bar style

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/320

Moves the app to Cheerio!


app/actions/Web.tsx
```diff
@@ -4,31 +4,28 @@ import {toCard} from './Card'
...
+const Cheerio = require('cheerio'); // Doesn't like import statements
 
 export function fetchQuestXML(id: string, url: string) {
   return (dispatch: Redux.Dispatch<any>): any => {
-    $.get(url, function(data: XMLElement | string) {
-      dispatch(loadQuestXML(id, data, defaultQuestContext()));
+    $.ajax({url,
+      dataType: 'text',
+      success: (data: string) => {
+        const quest = Cheerio.load(data)('quest');
```
smartin015: If the XML is malformed, this could throw an exception. Have we tested against this?

app/actions/Web.tsx
```diff
@@ -96,7 +93,7 @@ export function subscribe(email: string) {
       if (xhr.status === 200) {
         window.FirebasePlugin.logEvent('user_subscribe', email);
       } else {
-        window.FirebasePlugin.logEvent('user_subscribe_err', err);
+        // window.FirebasePlugin.logEvent('user_subscribe_err', xhr.status); // causes crash on iOS, also not critical
```
smartin015: p'haps we can just remove this line for now.

app/quests/test_quest.xml
```diff
@@ -5,7 +5,7 @@
     <choice text="End now">
       <trigger data-line="19">end</trigger>
     </choice>
-    <choice text="Fight">
+    <choice text="Fight one={{one}}">
```
smartin015: What is `one`?

toddmedema: It's equal to 1 - it's just there as a visual check that variable rendering is working (it is the test quest, after all :)

app/actions/Web.tsx
toddmedema: Fair. Been thinking about moving off Firebase / to Google Analytics anyway

app/actions/Web.tsx
toddmedema: Good idea. Just tried:
```
const foo = Cheerio.load('fsfidjsoj<fd></sf');
console.log(foo);
```
And no errors were thrown / foo was a valid Cheerio object. Seems like we're good against malformed XML, then (valid QDL XML is a separate issue....)

smartin015: It'd be nice to have this be a unit test, e.g. fake web response with bogus XML, to ensure we surface some error. Split this into #331.

Issue #142, opened 2016-12-02
Alerts: Indicate failed network requests to user


toddmedema:
When a search, download or Auth API request fails, notify the user that something went wrong

Use a snackbar: http://www.material-ui.com/#/components/snackbar

PR #335, opened 2017-05-19
add snackbar with network errors

toddmedema:
screen shot 2017-05-18 at 11 54 55 pm

Closes https://github.com/ExpeditionRPG/expedition-app/issues/142


app/reducers/Snackbar.tsx
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export function snackbar(state: SnackbarState = default_state, action: Redux.Action): SnackbarState {
+  switch(action.type) {
+    case 'SNACKBAR_OPEN':
+      const openAction = (action as SnackbarOpenAction);
+      return {
+        open: true,
+        message: openAction.message,
+        timeout: openAction.timeout || 4000,
```
smartin015: Does this mean `state.snackbar.open` will be open even after the snackbar has timed out?
If so, thoughts on just storing the message and a timestamp at which the snackbar closes?

app/actions/Web.tsx
```diff
@@ -59,10 +62,10 @@ export function search(numPlayers: number, user: UserState, search: SearchSettin
     // TODO: Pagination
     xhr.open('POST', authSettings.urlBase + '/quests', true);
     xhr.setRequestHeader('Content-Type', 'text/plain');
-    xhr.onload = function() {
+    xhr.onload = () => {
       var response: any = JSON.parse(xhr.responseText);
       if (response.error) {
-        throw new Error(response.error);
+        return dispatch(openSnackbar('Network error when searching: ' + response.error));
```
smartin015: What happens when `response.error` is enormous?

app/actions/Web.tsx
```diff
@@ -93,7 +99,7 @@ export function subscribe(email: string) {
       if (xhr.status === 200) {
         window.FirebasePlugin.logEvent('user_subscribe', email);
       } else {
-        console.log('Error encountered when subscribing: ' + err);
+        dispatch(openSnackbar('Error when subscribing: ' + err));
```
smartin015: Same question here.

app/actions/Web.tsx
```diff
@@ -128,10 +134,10 @@ export function submitUserFeedback(quest: QuestState, settings: SettingsType, us
       if (xhr.status === 200) {
         window.FirebasePlugin.logEvent('user_feedback_' + userFeedback.type, data);
         dispatch(userFeedbackClear());
-        return alert('Submission successful. Thank you!');
+        return dispatch(openSnackbar('Submission successful. Thank you!'));
       }
       window.FirebasePlugin.logEvent('user_feedback_' + userFeedback.type + '_err', data);
-      alert('Error encountered when submitting: ' + err);
+      return dispatch(openSnackbar('Error submitting feedback: ' + err));
```
smartin015: And here.

app/reducers/Snackbar.tsx
toddmedema: Nope - the snackbar material UI API fires the snackbar close event on timeout, which we then have set open to false

app/actions/Web.tsx
toddmedema: Just tested - truncates it to a single line

PR #339, opened 2017-06-02
show star ratings in search results

toddmedema:
Only appears if there's at least one rating:
screen shot 2017-06-02 at 4 08 41 pm


app/style.scss
```diff
@@ -324,8 +327,15 @@ body {
         }
       }
     }
+    .quantity {
+      height: 50px;
```
smartin015: Can these be constants?

app/components/Search.tsx
```diff
@@ -156,33 +157,20 @@ function renderResults(props: SearchProps): JSX.Element {
...
       <div className="searchDetails">
-        <h3>{props.selected.title}</h3>
-        <div className="author">by {props.selected.author}</div>
+        <h3>{quest.title}</h3>
+        <div className="author">by {quest.author}</div>
         <p>
-          {props.selected.summary}
+          {quest.summary}
         </p>
+        {quest.ratingcount && quest.ratingcount >= 1 && <StarRating readOnly={true} value={+quest.ratingavg} quantity={quest.ratingcount}/>}
```
smartin015: As implemented, one bad rating can sink your quest permanently, and one good rating can boost a quest past long-time best quests (like that scene from ender's game). I think this minimum rating should be higher... maybe 5? Possibly even a function of the variance of reviews?

toddmedema: Good idea - bumped up to 5 for now. We don't currently have a way to calculate variance (since we only store the count and avg with the quest) - I think we can look into that more if this is still an issue later

app/style.scss
toddmedema: Good point, added

app/components/base/StarRating.tsx
```diff
@@ -8,7 +8,8 @@ import StarBorder from 'material-ui/svg-icons/toggle/star-border'
 
 export interface StarRatingProps {
   hintText?: boolean;
-  onChange: (rating: number) => any;
+  onChange?: (rating: number) => any;
+  quantity?: number; // parens number to the right of the stars, aka quantity of ratings
```
smartin015: Let's remove this comment. It's dependent on UI implementation and doesn't do much to clarify.

PR #340, opened 2017-06-03
add content rating definitions

toddmedema:
These will be used by the quest creator, and by the app search filter


app/Constants.tsx
```diff
@@ -41,6 +41,28 @@ export const GENRES = [
...
+    nudity: 'No nudity allowed.',
+  },
+  Teen: {
+    violence: 'May contain brief, limited descriptions of violence.',
+    language: 'May contain some profanity.',
+    drugs: 'May contain drug use, but not abuse.',
+    nudity: 'May contain non-sexual nudity.',
+  },
+  Adult: {
+    violence: 'Violence allowed.',
```
smartin015: Prepend "All" to these? Otherwise it seems unclear how different one can make it from a Teen rating.

Issue #327, opened 2017-05-05
Tags, search and metadata upgrade


toddmedema:
@smartin015 for your review + our discussion this weekend

Mission: To make it easier for people to find quests based on a flexible set of criteria / tags (such as "funny", "family friendly", "horror", etc)

Mission: To get quests accurately tagged in a manner that's easy for quest writers and allows the addition of new tags as time goes on.

toddmedema:
New proposal:

now:
- playtime shortest should order by smallest MAX time, longest should order by largest MIN time
- search: move player count to top left, 7 quests for 1 <adventurer>, remove from search cards
search title:
text-overflow: ellipses
white-space: nowrap
overflow: hidden



Quest Creator:
- on publish, show modal asking for metadata, move all metadata except title
  - fetch from existing published quest, otherwise empty. Validate 
  - drop downs for genre, mood, content rating options (with explanations for content rating)
    - find list of genre and moods, such as from Plex + Amazon Video
  - also save metadata to shadow document, a la notes
  - v2: show preview of search result with metadata provided
- add warnings that quest metadata has been deprecated, now set as part of publishing

App: no longer pull metadata from XML, pull from quest DB object

sorts:
<existing>
- user rating
- show sort on property, i.e. default search of created on would show delta since date created, rating would show stars + # of ratings
  - on bottom of card, low in hierarchy

filters:
- text search input -> filling in creates overlay with live results section and slimmed down results showing just title and author, adds close / clear button to right of input, clicking result goes to details
- for adventurers, changeable in settings
- genre
- mood
- content rating
- play time min and max drop downs

details card:
- other quests by this author

end of quest card:
- other quests by this author

PR #341, opened 2017-06-03
improve search style; add min / max time, genre, content rating and highest rated sort

toddmedema:
closes https://github.com/ExpeditionRPG/expedition-app/issues/327

This will need to be matched with a QC update that adds support for the new sorts and filters, which I'll work on next

screen shot 2017-06-03 at 4 58 50 pm

screen shot 2017-06-03 at 2 45 51 pm


toddmedema:
Additional note / thought - probably should wait to roll this out till we have some quests tagged with these things, otherwise there'll be a big user rush to try it out and they'll get 0 results... 

So, let's roll out https://github.com/ExpeditionRPG/expedition-quest-creator/pull/277 this week, and this PR (perhaps the app in general?) next week

app/Constants.tsx
```diff
@@ -61,7 +61,7 @@ export const CONTENT_RATINGS = {
     drugs: 'All drugs allowed.',
     nudity: 'All nudity allowed.',
   },
-};
+} as any; // allows using CONENT_RATINGS[missing_key] for empty / default states
```
smartin015: `export const CONTENT_RATINGS: {[key: string]: {violence: string, language: string, drugs: string, nudity: string} = ...` (or something similar to that)

Also nit: `CONENT` in comment.

Also also, why don't we case on the empty state elsewhere instead of relying on a null lookup?

app/components/Search.tsx
```diff
@@ -52,48 +53,126 @@ class SearchSettingsCard extends React.Component<SearchSettingsCardProps, {}> {
...
+            floatingLabelFixed={true}
+            onChange={(e: any, i: any, v: string) => this.onChange('mintimeminutes', v)}
+            value={this.state.mintimeminutes}
+            style={{color: 'black'}}
+            floatingLabelStyle={{color: 'black'}}
+            iconStyle={{fill: 'black'}}
+            underlineStyle={{borderColor: 'black'}}
+            selectedMenuItemStyle={{paddingRight: '50px'}}
+          >
+            <MenuItem value={null} primaryText="Any length"/>
```
smartin015: Prefer this (and max time) to be generated so min and max time won't accidentally drift

app/reducers/StateTypes.tsx
```diff
@@ -43,6 +43,10 @@ export interface SearchSettings {
   age: string;
   order: string;
   owner: string;
+  mintimeminutes: number;
+  maxtimeminutes: number;
+  contentrating: string;
```
smartin015: Seems like contentrating and genre should be enums.

toddmedema: Good idea, added

app/Constants.tsx
toddmedema: Improved the typescript definition...

Learning time! Which made me realize I had the wrong description of why it's broken. It's not that it breaks on the empty case... it's because the UI generates the current description text by iterating over the keys for maintainability: `{Object.keys(rating).map((key: string, i: number) => {` (aka violence, drugs, etc)

Which, if you try to define CONTENT_RATINGS as explicitly having the keys violence, drugs, etc, throws this typescript error: `Element implicitly has an 'any' type because type 'ContentRatingType' has no index signature.`

So, then, you'd think you could create an index signature using a type - that'd be easy! Create an enum for violence, drugs, etc, and type it as [key: LabelType]: string..... but, NOPE, typescript only lets you use `string` and `number` for index signatures. So, the best we can do if we want the more maintainable for-loop generation (as far as I can tell) is `export type ContentRatingType = {[key: string]: {[key: string]: string}};`, which is still decent and good enough for now IMHO

app/components/Search.tsx
toddmedema: Done

toddmedema:
UI updated, added the details snippet at the bottom of results based on order:

screen shot 2017-06-08 at 11 03 23 pm
screen shot 2017-06-08 at 11 04 28 pm

app/Constants.tsx
```diff
@@ -32,16 +32,20 @@ export const REGEX = {
...
+
+export type GenreType = 'Comedy' | 'Drama' | 'Horror' | 'Mystery' | 'Romance';
 export const GENRES = [
   'Comedy',
   'Drama',
   'Horror',
   'Mystery',
   'Romance'
-];
+] as [GenreType];
```
smartin015: `export const GENRES: GenreType[]`. Since type assertion is dumb.

Ditto below.

app/actions/Web.tsx
```diff
@@ -57,9 +57,21 @@ export function search(numPlayers: number, user: UserState, search: SearchSettin
     if (search.order) {
       params.order = search.order;
     }
+    if (search.mintimeminutes) {
```
smartin015: I feel like we could use spreading to make this less boilerplatey.

app/components/Search.tsx
```diff
@@ -53,48 +55,127 @@ class SearchSettingsCard extends React.Component<SearchSettingsCardProps, {}> {
     this.setState({[attrib]: value});
   }
 
+// TODO once Material UI adds support for theming SelectFields (https://github.com/callemall/material-ui/issues/7044)
+  // then remove the clutter here / move to Theme.tsx
```
smartin015: Nit: same-line comments

app/components/Search.tsx
```diff
@@ -53,48 +55,127 @@ class SearchSettingsCard extends React.Component<SearchSettingsCardProps, {}> {
...
   render() {
+    const rating = CONTENT_RATINGS[this.state.contentrating];
+    const timeBuckets = PLAYTIME_MINUTES_BUCKETS.map((minutes: number, index: number) => {
+      return <MenuItem key={index} value={minutes} primaryText={`${minutes} min`}/>;
+    });
+    // TODO Until we have at least 3 quests in all genres, only show limited options
+    const visibleGenres = ['Comedy', 'Drama'] as GenreType[];
```
smartin015: Use type assignment, not type assertion (`const visibleGenres: GenreType[]`)

app/Constants.tsx
toddmedema: Done

app/actions/Web.tsx
toddmedema: Truth! Aligned UI with API fields, was able to reduce to:
```
let params: any = {
      players: numPlayers,
      ...search,
    };
    Object.keys(params).forEach((key: string) => { if (params[key] === null) { delete params[key]; }});
```
With only a small update to API: https://github.com/ExpeditionRPG/expedition-quest-creator/commit/80cc93932f80166fa4639ea7a882cfafc9a073d1

app/components/Search.tsx
toddmedema: Done

app/components/Search.tsx
toddmedema: Done

toddmedema:
screen shot 2017-06-10 at 10 36 56 am


screen shot 2017-06-10 at 9 56 40 am

toddmedema:
@smartin015 ready for re-review

app/Constants.tsx
```diff
@@ -42,7 +44,9 @@ export const GENRES = [
 ];
 
 // Content rating options and their definitions, generally based on MPAA guidelines
-export const CONTENT_RATINGS = {
+export type ContentRatingLabelType = 'Everyone' | 'Teen' | 'Adult';
+export type ContentRatingType = {[key: string]: {[key: string]: string}};
```
smartin015: Have you tried `{[key: ContentRatingLabelType]: {[key: 'violence' | 'language' | 'drugs' | 'nudity']: string}};`?

app/actions/Web.tsx
```diff
@@ -43,27 +43,18 @@ export function search(numPlayers: number, user: UserState, search: SearchSettin
...
-    if (search.text) {
-      params.search = search.text;
-    }
-    if (search.age && search.age !== 'inf') {
-      params.published_after = parseInt(search.age, 10);
-    }
-    if (search.order) {
-      params.order = search.order;
-    }
+    let params: any = {
```
smartin015: Nit: Could probs single-line this assignment

app/actions/Web.tsx
```diff
@@ -43,27 +43,18 @@ export function search(numPlayers: number, user: UserState, search: SearchSettin
...
+      players: numPlayers,
+      ...search,
+    };
+    Object.keys(params).forEach((key: string) => { if (params[key] === null) { delete params[key]; }});
 
-    var xhr = new XMLHttpRequest();
-    // TODO: Pagination
+    let xhr = new XMLHttpRequest();
```
smartin015: `const`?

app/Constants.tsx
toddmedema: Yep, Typescript doesn't let you do that: `An index signature parameter type must be 'string' or 'number'.`

PR #344, opened 2017-06-05
Add Google Analytics to web app

toddmedema:
Merging now / for post review

As discussed IRL, Firebase doesn't currently work on web app, and Google Analytics doesn't currently run on Android / iOS, so this at least gets us the ability to run basic queries like how many people are using web vs Android vs iOS


app/index.html
```diff
@@ -55,5 +55,18 @@
     <p>The Companion App for <a href="https://expeditiongame.com">Expedition: The Roleplaying Card Game</a>.</p>
     <p><b>JavaScript is currently disabled. You must enabled JavaScript to use the app and play Expedition: The RPG.</b></p>
   </noscript>
+  <script>
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
```
smartin015: Does this throw error on mobile? Can we gate this on web-only?

toddmedema: Wasn't throwing an error, but good point - no need to load it if it we aren't using it!

PR #345, opened 2017-06-05
pull quest info from search results / remove dependency on XML metadata

toddmedema:


app/actions/Quest.test.tsx
```diff
@@ -1,4 +1,3 @@
-
 describe('Quest action', () => {
-  it('TODO');
+  it('Init request successfully returns the parsed quest node');
```
smartin015: Prefer:
```javascript
describe('initQuest', () => {
  it('successfully returns the parsed quest node');
});
```

Basically, use `describe` to pare down until you get to the thing you want to test, then use `it` to test individual conditions of that thing. [src](https://stackoverflow.com/questions/32055287/what-is-the-difference-between-describe-and-it-in-jest)

app/actions/Web.tsx
```diff
@@ -6,33 +6,32 @@ import {initQuest} from './Quest'
...
-export function fetchQuestXML(id: string, url: string) {
+export function fetchQuestXML(details: QuestDetails) {
   return (dispatch: Redux.Dispatch<any>): any => {
-    $.ajax({url,
+    $.ajax({url: details.publishedurl,
       dataType: 'text',
       success: (data: string) => {
-        const quest = Cheerio.load(data)('quest');
-        dispatch(loadQuestXML(id, quest, defaultQuestContext()));
+        const node = Cheerio.load(data)('quest');
```
smartin015: I've been trying to keep `node` to be used for `ParserNode` instances. Maybe put `xml` here or some such?

app/actions/Web.tsx
```diff
@@ -6,33 +6,32 @@ import {initQuest} from './Quest'
...
+        dispatch(loadQuestXML(details, node, defaultQuestContext()));
       },
       error: (xhr: any, error: string) => {
-        dispatch(openSnackbar('Network error when fetching quest'));
+        dispatch(openSnackbar('Network error, please check your connection'));
```
smartin015: Nit: Add period and use semicolon or hyphen (also below)

app/actions/Quest.test.tsx
toddmedema: Gotcha

app/actions/Web.tsx
toddmedema: Changed to questNode, since that's what it's called in all of the function arguments down the line (though I could also see that getting changed to questXml if you don't want `node` in there at all)

app/actions/Web.tsx
toddmedema: Done

PR #347, opened 2017-06-06
Quest crawler implementation

smartin015:
Lets us do "compile-time" checking of runtime errors, and generating stats on quest length etc.

A couple tangential changes also present:
* ParserNode now has `getVisibleKeys` and `getComparisonKey` to allow for edge following and "visited" checking.
* Context now has a `path` attribute that will be used later for node reconstruction (if we go to a service model).

### DIFF
--- a/app/parser/Crawler.tsx
+++ b/app/parser/Crawler.tsx
@@ -15,9 +15,8 @@ type DoFn = (s: CrawlerStats, n :ParserNode) => void;
 
 
 export class Crawler {
-  // TODO: Privatize
-  public statsById: {[id:string]: CrawlerStats};
-  public statsByLine: {[line:number]: CrawlerStats};
+  statsById: {[id:string]: CrawlerStats};
+  statsByLine: {[line:number]: CrawlerStats};
   seen: Set<string>;
 
   constructor() {
@@ -79,15 +78,6 @@ export class Crawler {
     let maxPath = 0;
     while(queue.length > 0 && maxPath < 50) {
       let q = queue.shift();
-
-      // This happens when we hit the end of a quest.
-      if (q.node.getTag() === 'trigger' && q.node.elem.text().trim() === 'end') {
-        console.log('end trigger seen');
-        this.statsById[q.prevId].causeOf.add('END');
-        this.statsByLine[q.prevLine].causeOf.add('END');
-        continue;
-      }
-
       const id = q.node.elem.attr('id') || q.prevId;
       const line = parseInt(q.node.elem.attr('data-line'), 10);
       console.log('Working in node ' + id + ' line ' + line);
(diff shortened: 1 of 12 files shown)


app/parser/Crawler.tsx
```diff
@@ -0,0 +1,180 @@
...
+    this.seen = new Set();
+    this.statsById = {};
+    this.statsByLine = {};
+  }
+
+  public crawl(root: ParserNode) {
+    this.traverse(root);
+  }
+
+  // Retrieves stats for a given node ID.
```
toddmedema: Interesting... so, is this how we'd expose things to quest writers, on a collapsed node basis? Seems easier to keep track of than on a per line or per card basis, given that those are constantly changing - we'll just need to explain this to the user

app/parser/Crawler.tsx
```diff
@@ -0,0 +1,180 @@
+import {ParserNode} from './Node'
+import {handleAction} from './Handlers'
+
+export interface CrawlerAction {}
+
+export interface CrawlerStats {
+  resultOf: Set<string>;
+  causeOf: Set<string>;
```
toddmedema: I keep mentally hiccuping on this / confusing is with `resultOf`... would `causes` make more sense?

toddmedema: Though, that runs into the same problem... basically, it's not 110% clear that causes is the output of the node, and not the list of things that cause the node. What about simplifying them to be `inputs` and `outputs`?

app/parser/Crawler.tsx
```diff
@@ -0,0 +1,180 @@
...
+  minPathActions: number;
```
toddmedema: path actions is one thing, but each action takes wildly different amounts of time (ie one sentence of text w/ next, vs two paragraphs with multiple choices, vs tier 1 combat, vs tier 6 combat).

Is this the code that'll programmatically calculate / estimate quest playtimes? If so, think we should also have `minPathMinutes` and `maxPathMinutes` (then we could also have separate estimates for the minimum and maximum duration for each of those above cases, ie a two paragraph roleplay with two choices could take minimum of 0.5 minutes and maximum of 2 minutes)

app/parser/Node.tsx
```diff
@@ -15,14 +15,40 @@ function getNodeAttributes(e: CheerioElement): {[key:string]:string;} {
   return e.attribs || (e.get(0) as any as CheerioElement).attribs;
 }
 
+function getChildNumber(domElement: DOMElement): number {
+    var i=1;;
```
toddmedema: extra ;

app/reducers/QuestTypes.tsx
```diff
@@ -1,4 +1,4 @@
-
+import {encounters} from '../Encounters'
```
toddmedema: Doesn't look like this is used anywhere in this file

app/parser/Crawler.tsx
smartin015: Nope, that'll be tacked on later. I wanted a base implementation that just traverses the quest correctly so we can extend it later for multiple functions (including time calculation).

app/parser/Crawler.tsx
smartin015: Ideally, it'd be both. I figured the collapsed node format would be how we'd present more aggregate "Graph" views of the quest - whether that's with a visual graph or via collapsed directory structure (or something else entirely). But we should still be surfacing errors as close to where they occur as possible.

Added benefit of this approach: gotos are probably the hardest to keep track of mentally, but since they always occur on an ID boundary we can clearly track those edges with this approach.

app/parser/Node.tsx
smartin015: I just reeeealy wanted to set i equal to 1 :P

app/reducers/QuestTypes.tsx
smartin015: Good catch, fixed.

app/parser/Crawler.tsx
smartin015: Yeah, inputs/outputs does make a lot more sense. Naming is hard.

app/parser/Crawler.tsx
toddmedema: SGTM

app/parser/Crawler.tsx
toddmedema: SGTM

PR #349, opened 2017-06-10
Simplified crawler, removing stats generation

smartin015:
The crawler in the app is now an abstract class, where the inheriting class implements `onNode` and `onEvent` handlers to construct some kind of meaning from the traversal. This makes the graph traversal code way simpler, and allows us to keep stats generation code in the IDE repo where it belongs.

### DIFF
--- a/app/parser/Crawler.test.tsx
+++ b/app/parser/Crawler.test.tsx
@@ -1,4 +1,4 @@
-import {Crawler} from './Crawler'
+import {CrawlerBase, CrawlEntry, CrawlEvent} from './Crawler'
 import {ParserNode} from './Node'
 import {defaultQuestContext} from '../reducers/QuestTypes'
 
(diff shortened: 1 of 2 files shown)


app/parser/Crawler.test.tsx
```diff
@@ -74,28 +99,45 @@ describe('Crawler', () => {
...
+        expect(e).not.toEqual('END');
+        expect(e).not.toEqual('INVALID');
+      }, null);
       crawler.crawl(new ParserNode(xml, defaultQuestContext()));
 
-      expect(Array.from(crawler.getStatsForId('A1').outputs)).toEqual(['IMPLICIT_END']);
+      expect(foundImplicitEnd).toEqual(true);
     });
 
     it('safely handles nodes without line annotations', () => {
```
toddmedema: It seems like lack of line annotations would break a great many things down the crawling road (for example, here it's unable to finish crawling the quest).

I'd be in favor of it erroring / stopping / alerting in this case, because something's clearly gone wrong and any attempt to continue crawling will just lead to nonsense results.... (like saying that your quest only takes 20 seconds to play)

smartin015: The neat part about this being an abstract class is that we can handle this however we want - if we stop and error on an invalid node, that means we won't be able to find and show any other invalid nodes the user may have in their quest.

We should definitely treat INVALID as strong errors in the IDE, but I'd prefer the logic that prevents stats calculation stuff to happen in the inheriting class. (Added a TODO in the IDE impl)

smartin015:
Merging after offline discussion.

PR #355, opened 2017-06-11
Action tests + login cleanup

toddmedema:


app/actions/User.tsx
```diff
@@ -1,94 +1,97 @@
...
 import {UserState} from '../reducers/StateTypes'
 import {authSettings} from '../Constants'
 
 declare var gapi: any;
 declare var window: any;
 
 
-function registerUserAndIdToken(user: {name: string, image: string, email: string}, idToken: string, cb: (user:UserState) => any) {
-  var xhr = new XMLHttpRequest();
+function registerUserAndIdToken(user: {name: string, image: string, email: string}, idToken: string, callback: (user:UserState, err?: string) => any) {
```
smartin015: We could actually `type UserLoginCallback = (user: UserState, err?: string) => any;`, which would give us a standard callback interface for all of these things (and reduce redundant code).

app/actions/User.tsx
```diff
@@ -122,17 +128,17 @@ export function logout() {
 }
 
 /*
```
smartin015: hmmmmmmmmm....

app/actions/User.tsx
toddmedema: Good idea

app/actions/User.tsx
toddmedema: Getting logout to work's outside of the scope of this PR, but I'd bet that the commented-out code'll be pretty close to the final solution / don't want to delete it

smartin015: Fair 'nuff.

PR #360, opened 2017-06-14
React.tsx refactor

smartin015:
Attempting to make the root tsx file more testable/reliable. This splits each part of the setup into a function, and adds a level of indirection between the functions and the global `document`/`window`/`device`/etc objects. 

Also includes some small changes to related code, and moves google analytics inside typescript land.

### DIFF
--- a/app/React.tsx
+++ b/app/React.tsx
@@ -1,32 +1,5 @@
 declare var require: any;
 declare var module: any;
-declare var device: any;
-declare var gapi: any;
-declare var ga: any;
-
-export interface Document {
-  body: {className: string};
-  addEventListener: (e: string, f: ()=>any, useCapture?: boolean) => void;
-  getElementById: (id: string) => Element;
-}
-
-export interface ReactWindow extends Window {
-  platform?: string;
-  APP_VERSION?: string;
-  AndroidFullScreen?: {
-    immersiveMode: (success: () => any, failure: () => any) => void,
-  };
-  FirebasePlugin?: {
-    onTokenRefresh?: (success: (token: string) => any, failure: (error: string) => any) => void,
-    logEvent: (name: string, args: any) => any,
-  };
-  plugins?: {
-    insomnia: {keepAwake: ()=>void},
-  };
-  test?: boolean;
-  device?: {platform: string};
-}
-declare var window: ReactWindow;
 
 import * as React from 'react'
 import * as ReactDOM from 'react-dom'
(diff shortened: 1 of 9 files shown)


app/React.tsx
```diff
@@ -1,109 +1,149 @@
...
-// Wait for device API libraries to load
-document.addEventListener('deviceready', onDeviceReady, false);
+const PACKAGE = require('../package.json');
+const injectTapEventPlugin = require('react-tap-event-plugin');
 
-// device APIs are available
-window.platform = 'web';
-function onDeviceReady() {
-  var p = device.platform.toLowerCase();
+export function getAppVersion(): string {
```
toddmedema: Why are these functions (getting app version, platform, logging, etc) in the main file instead of globals? Given that they're used in other files, I feel like globals would be a better place for them

app/React.tsx
```diff
@@ -112,11 +152,70 @@ let render = () => {
...
+    return;
+  }
+  setGA(ga);
+  ga('create', 'UA-47408800-9', 'auto');
+  ga('send', 'pageview');
+  console.log('google analytics set up');
+}
+
+export function init() {
+  // TODO: remove these and have everyone use getPlatform()/getVersion().
```
toddmedema: seems like this comment is done? Or that the change would be small + fast enough to be worth including in this PR since it's just an import statement away

app/components/base/Card.tsx
```diff
@@ -42,15 +43,21 @@ export default class ExpeditionCard extends React.Component<ExpeditionCardProps,
...
+            break;
+          case 'web':
+            throw new Error('Cannot rate web app');
+          default:
+            throw new Error('Uknown platform encountered');
         }
         break;
       case 'FEEDBACK':
-        window.open(URLS.feedback, '_system');
+        window.open(URLS.feedbackBase + getDevicePlatform(), '_system');
```
toddmedema: Seeing this in implementation, I wonder if it should be getFeedbackUrl(), which returns `feedbackBase + getDevicePlatform()`, since it seems like this is the only way we'd generate feedback URLs (since the base url explicitly ends in `?platform=` / it'd break if we did anything else). If we wanted to add support for other params in the future, we could add the ability to pass a dict into `getFeedbackUrl(urlParams)` that'd query string them.

app/React.tsx
smartin015: Fix'd.

app/components/base/Card.tsx
smartin015: Seems jinky to have formatting functions in a file labeled `Constants.tsx`. Let's keep it as a private class method at most and revisit if we ever use it elsewhere.

Issue #365, opened 2017-06-19
Visual feedback when successfully tap number counter


toddmedema:
Right now you can't actually tell if you tapped it or not, unless you happen to also be looking at the number

PR #369, opened 2017-06-25
picker ui: increase size and add ripple effect

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/365

Turns out, IconButtons don't do nearly a good of job at ripples as Flat Buttons, and since you can assign an icon to FlatButtons, they're effectively the same thing... As much as I love Material-UI, they do a really bad job of having feature parity across their elements....

Before and after: (also on beta if you want to see it live)
screen shot 2017-06-25 at 10 59 35 am
screen shot 2017-06-25 at 10 59 29 am


toddmedema:
The background change is only if it's "active", aka it's the last input you touched. I think that improves usability (even above the touch ripple effect, because it doesn't require a quick reaction speed to see the response). LMK if that seems reasonable

PR #372, opened 2017-06-27
change app to point to `api.`

toddmedema:
Instead of `quests.`

However, doesn't seem to be working quite yet... not sure if DNS hasn't fully propagated, or if the Google Auth stuff doesn't like redirects...


smartin015:
What's the error? I could see the allowed gapi redirects list needing to be changed in the cloud console.

toddmedema:
Turns out this was a flaw in the S3 bucket redirect system. Changed it so that the heroku instance actually exists on both domains, and now it works (including a few other fixes / tweaks)

README.md
```diff
@@ -1,7 +1,9 @@
-# The Expedition Roleplaying Card Game App
+# Expedition: The Companion App
```
smartin015: On the android store this is listed as "Expedition: The RPG Card Game".

README.md
```diff
@@ -1,7 +1,9 @@
...
 
 [![Build Status](https://travis-ci.org/ExpeditionRPG/expedition-app.svg)](https://travis-ci.org/ExpeditionRPG/expedition-app)
 
+Learn more about Expedition and play the game at [ExpeditionGame.com](https://expeditiongame.com). The app is available at [App.ExpeditionGame.com](https://App.ExpeditionGame.com), and for Android and iOS.
```
smartin015: Link to web play store & apple store?

app/Globals.tsx
```diff
@@ -5,7 +5,7 @@ declare var ga: any;
 const PACKAGE = require('../package.json');
 
 export interface ReactDocument extends Document {
-  addEventListener: (e: string, f: ()=>any, useCapture?: boolean) => void;
```
smartin015: Why?

app/actions/User.tsx
```diff
@@ -11,23 +11,34 @@ type UserLoginCallback = (user: UserState, err?: string) => any;
...
       loggedIn: true,
-      id: xhr.responseText,
+      id,
       name: user.name,
       image: user.image,
       email: user.email,
     });
-  };
-  xhr.onerror = () => {
+  }).catch((error: string) => {
```
smartin015: Are you sure this is `string` and not `Error`? If `Error`, we're probably better off just logging it to console alone.

package.json
```diff
@@ -5,6 +5,15 @@
     "homepage": "http://expeditiongame.com",
     "bugs": "https://github.com/ExpeditionRPG/expedition-app/issues",
     "license": "CC-BY-NC-ND-4.0",
+    "scripts": {
+        "test": "mocha meta_tests.js && karma start --single-run --browsers PhantomJS",
+        "watch:test": "karma start",
+        "dev": "webpack-dev-server --progress",
+        "build": "webpack --config ./webpack.dist.config.js --progress"
+    },
+    "engines": {
```
smartin015: engines lives at the bottom of the quest creator `package.json`

project.sh
```diff
@@ -216,30 +216,6 @@ if [[ $init = 1 ]] || [[ $icons = 1 ]] ; then
 fi
 
 # ----
-# Add Plugins
-
-if [[ $init = 1 ]] || [[ $plugins = 1 ]] ; then
```
smartin015: Nice, glad to see this can be replaced by node modules.

app/actions/User.tsx
toddmedema: It's an error. I find it useful to still log the context of errors in the console, so that if there's multiple error messages happening you can easily tell where in the code each is occuring

package.json
toddmedema: I prefer everything at the top, so that it's everything small and important, and then the huge list of dependencies. I guess the main alternative would be alphabetical order, but that's already violated by having name at the top, which makes sense to me *shrugs*

app/Globals.tsx
toddmedema: Typescript complained about it, but the error message was so long that it truncated it before it told me what it expected to see. Guessing that it's something from a new version of Typescript.

PR #384, opened 2017-07-11
reduce 1 & 2 player damage taken

toddmedema:
Seems like we've gotten several feedback recently that 1 and 2 player combat is still just a touch too difficult. This reduces the 1 player multiplier from 0.5 to 0.4, 2 player from 0.75 to 0.6, 3 player from 0.9 to 0.8.

Given that these numbers get rounded at the end, it shouldn't result in too big a difference, but should help a bit :) `damage = Math.min(10, Math.round(damage * combat.damageMultiplier * playerMultiplier));`


app/cardtemplates/combat/Actions.tsx
```diff
@@ -63,7 +63,13 @@ function getEnemies(node: ParserNode): Enemy[] {
...
   // general balance based on 4 players, scaling up / down on a curve
   // since a bit more or less damage makes a huge difference in # of rounds survivable
-  const playerMultiplier = Math.sqrt(settings.numPlayers) / 2;
+  // 1 = 0.4
+  // 2 = 0.6
+  // 3 = 0.8
+  // 4 = 1
+  // 5 = 1.1
+  // 6 = 1.2
+  const playerMultiplier = Math.pow(settings.numPlayers, 1.8) / 2 - 0.1;
```
smartin015: With `settings.numPlayers = 2`, this evaluates to 1.64...

Fitting these to a curve via quadratic regression gives: `0.13 + 0.27535714x - 0.016071429x^2`. with an accuracy of ~0.01. Perhaps we could use that instead?

toddmedema:
@smartin015 simplified to a constant dictionary-based implementation, LG?

PR #388, opened 2017-07-15
Interstitial roleplay during combat

smartin015:
This allows for people to specify an `<event on="round" if="..."/>` element within combat that is conditionally followed after every timer card (preempted by surge, but still shown).

As part of this PR, there's a minor extension of behavior with respect to the `on=` attribute. Any events dispatched via trigger (e.g. `**end**`) may be "caught and handled" by the nearest parent element with a matching `on=` attribute. 

What this practically means in the combat interstitial is that you can win and lose combat simply with `**win**` or `**lose**`.


app/cardtemplates/combat/Combat.test.tsx
```diff
@@ -1,192 +1,2 @@
-
 describe('Combat', () => {
-  it('TODO');
 });
-/*
```
toddmedema: Wouldn't it be helpful to keep these around as stubs?

app/parser/Context.tsx
```diff
@@ -91,7 +91,22 @@ export function updateContext(node: Cheerio, ctx: QuestContext, action?: string|
   }
 
   const nodeId = node.attr('id');
+
+  // Special handling of roleplay node - this is readonly and cannot be cloned.
+  let tmpCombatRoleplay: any = null;
+  if (ctx.templates && ctx.templates.combat && ctx.templates.combat.roleplay) {
+    tmpCombatRoleplay = ctx.templates.combat.roleplay
```
toddmedema: missing ;

app/parser/Context.tsx
```diff
@@ -91,7 +91,22 @@ export function updateContext(node: Cheerio, ctx: QuestContext, action?: string|
...
+    ctx.templates.combat.roleplay = null;
+  }
+
   let newContext: QuestContext = Clone(ctx);
+
+  // Reassign readonly (uncopyable) attributes
+  if (tmpCombatRoleplay) {
+    newContext.templates.combat.roleplay = tmpCombatRoleplay.clone();
```
toddmedema: Does .clone behave differently from Clone()? / Why don't we do .clone for the whole thing?

app/cardtemplates/combat/Combat.test.tsx
smartin015: `Combat.tsx` now exclusively handles rendering; these tests include everything from combat timer checks to player settings, to back-button logic using the old polymer libraries. This is already captured in other tests.

app/parser/Context.tsx
smartin015: Yep, `.clone` is a `ParserNode` method that knows about context and (some) readonly elements, while `Clone()` is a naive deep copy. QuestContext is an object and not a class with methods, so we don't have an equivalent `.clone` for it.

app/parser/Context.tsx
smartin015: Fixed.

toddmedema:
Plz fix merge conflicts, then feel free to merge

toddmedema:
From IRL, looks like there are still some bugs here that we'll want to resolve before pushing out, but hopefully once it's ready, the quest creator will be able to pick it up without any extra work

PR #389, opened 2017-07-15
Refactor unit test stubs and assign some bounties

smartin015:
These and future bountied tests should be searchable in code with `// $`


app/React.test.tsx
```diff
@@ -89,8 +89,8 @@ describe('React', () => {
   });
 
   describe('logEvent', () => {
-    it('logs to firebase');
-    it('logs to google analytics if GA set up');
+    it('logs to firebase'); // $10
```
toddmedema: This should be "logs to firebase if firebase set up"?

Issue #392, opened 2017-07-27
Proposal: 1+2 player rules improvement


toddmedema:
Solo play: Control 2 adventurers (aka squad mode); the app gives you 2x as much time each round. Difficulty / damage is set to the same as 2 player.
Solo + 2 player: draft pick your abilities (ie draw 3, keep 1)

Played it a few times myself, seems to address both the single player issue of not doing enough damage / ability diversity, and the 1-2 player issue of drawing crappy abilities and then being totally blocked b/c nobody has damaging abilities for entire rounds.

toddmedema:
While we continue to discuss and playtest internally, posted to the community on Reddit to get their feedback: https://www.reddit.com/r/expeditionrpg/comments/6qc7ro/new_solo_player_rules_beta/

PR #395, opened 2017-07-30
improve solo + 2 player rules

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/392


app/Constants.tsx
```diff
@@ -38,17 +38,27 @@ export const REGEX = {
 
 export const PLAYTIME_MINUTES_BUCKETS = [20, 30, 45, 60, 90, 120];
 
-// general balance based on 4 players, scaling up / down on a curve
+// Based on 4 players, scaling up / down on a curve
 // since a bit more or less damage makes a huge difference in # of rounds survivable
 export const PLAYER_DAMAGE_MULT: {[key: number]: number} = {
-  1: 0.4,
+  1: 0.6,
   2: 0.6,
```
smartin015: I think we should actually take 1 & 2 player damage down even farther (0.5?) - tried 2P yesterday, and we died 4x in combat with a tier 4 before finally beating it.

app/components/QuestStart.tsx
```diff
@@ -16,28 +16,38 @@ export interface QuestStartProps extends QuestStartStateProps, QuestStartDispatc
...
       </ol>
 
       {multiPlayer && <p>
-        During your adventure, pass this device to your right whenever you make
-        a decision or hit the "Next" button during the story.
+        During your adventure, pass this device to your right whenever you make a decision
+        or hit the "Next" button during the story.
         {props.settings.multitouch && <span> During combat, place this device in the center of the table.</span>}
-        {!props.settings.multitouch && <span> During combat, one player will manage the device.</span>}
+        {!props.settings.multitouch && <span> During combat, one player should manage the device.</span>}
```
smartin015: Can probably specify this only when multiplayer as well (since if there's only one player, they'll definitely be managing the device).

app/Constants.tsx
toddmedema: Good idea

app/components/QuestStart.tsx
toddmedema: This whole block is actually wrapped in a multiplayer-only check (Github diff might've cut it off)

PR #399, opened 2017-08-06
Expose round and surge details via op functions, fix op function binding

smartin015:
This exposes `_.currentCombatRound()` and `_.isCombatSurgeRound()`, which can be used to conditionally show/hide the new `* on round` combat event. This branch is also now followed after surge cards.

Added a few extra bits to the test quest to show this off.

### DIFF
--- a/app/cardtemplates/combat/Combat.tsx
+++ b/app/cardtemplates/combat/Combat.tsx
@@ -145,7 +145,6 @@ function renderSurge(props: CombatProps): JSX.Element {
       </span>
     );
   }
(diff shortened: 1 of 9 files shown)


app/cardtemplates/combat/Combat.tsx
```diff
@@ -144,6 +145,7 @@ function renderSurge(props: CombatProps): JSX.Element {
       </span>
     );
   }
+  console.log(props.node.ctx.templates.combat.roundCount);
```
toddmedema: Log

PR #402, opened 2017-08-13
Add announcement bar

toddmedema:
Would be good to roll this out by Tuesday at the latest so that it's live & on most devices for KS launch day

screen shot 2017-08-13 at 5 09 08 pm

Pairs with https://github.com/ExpeditionRPG/expedition-api/pull/11


app/Constants.tsx
```diff
@@ -1,9 +1,9 @@
 declare var window:any;
 
 export const authSettings = {
-  urlBase: 'https://api.expeditiongame.com',
+  // urlBase: 'https://api.expeditiongame.com',
   // urlBase: 'http://devquests.expeditiongame.com',
-  // urlBase: 'http://localhost:8081',
+  urlBase: 'http://localhost:8081',
```
smartin015: Fix dis

app/Style.scss
```diff
@@ -693,6 +693,28 @@ body {
         font-size: fontsize(flavor) !important;
       }
     }
+
+    &.announcing .logo {
+      // 2 lines of text @ 1.2 leading + small top + bottom padding for button + top margin for logo
```
smartin015: Even with the comment, I'd never want to touch this in a code change.... make constants?

app/actions/Announcement.tsx
```diff
@@ -0,0 +1,9 @@
+import {AnnouncementCloseAction, AnnouncementOpenAction} from './ActionTypes'
+
+export function closeAnnouncement(): AnnouncementCloseAction {
```
smartin015: Could halve the code needed here and remove an interface type by just adding a bool to `openAnnouncement` and change it to `setAnnouncement` or similar.

app/actions/Web.tsx
```diff
@@ -15,6 +16,22 @@ declare var window:any;
...
+  return (dispatch: Redux.Dispatch<any>): any => {
+    $.ajax({url: authSettings.urlBase + '/announcements',
+      dataType: 'json',
+      success: (data: any) => {
+        if (data.message !== null) {
+          dispatch(openAnnouncement(data.message, data.link));
+        }
+      },
+      error: (xhr: any, error: string) => {
+        // Don't alert user - it's not important to them if this fails
```
smartin015: But we should still be logging it to firebase, no?

app/components/SplashScreenContainer.tsx
```diff
@@ -3,14 +3,24 @@ import { connect } from 'react-redux'
...
+  return {
+    announcement: state.announcement,
+  };
 }
 
 const mapDispatchToProps = (dispatch: Redux.Dispatch<any>, ownProps: any): SplashScreenDispatchProps => {
   return {
+    onAnnouncementTap: (announcement: AnnouncementState) => {
+      if (announcement.link !== '') {
+        window.open(announcement.link + '?utm_source=app', '_system');
```
smartin015: Does this work on all versions (ios, android, web)? IIRC we had some weirdness with this earlier.

smartin015:
Also, tests failed?

app/Style.scss
toddmedema: Added in the line height constant; everything else is specific to this element

app/Constants.tsx
toddmedema: Done

app/actions/Announcement.tsx
toddmedema: That feels weird to me... the code that says "close the announcement" shouldn't need to know the default state, and having an ANNOUNCEMENT_SET action with two entirely different pieces of logic based on an open/close boolean seems like the definition of two functions to me?

app/actions/Web.tsx
toddmedema: True!

toddmedema:
Tests fail because main.tsx now calls a file (actions/Web) that references jQuery. Tried some quick hacks to get the test to pass, but none worked; we should switch off jQuery at some point any way, but that's far outside the scope of this and whatever new library we use will require a lot of testing on different devices

app/components/SplashScreenContainer.tsx
toddmedema: Just tried on web + android + iOS. Works!

toddmedema:
@smartin015 applied code review changes and responded to comments

app/actions/Announcement.tsx
smartin015: It doesn't have to, and the `ANNOUNCEMENT_SET` action only has a single piece of logic, which is to set the state of the announcement pane.

`setAnnouncement(false)`

and 

`setAnnouncement(true, "message", "link");`

can both call the same function `function setAnnouncement(open: bool, message?: string, link?: string)`.

app/actions/Web.tsx
```diff
@@ -15,6 +16,23 @@ declare var window:any;
 declare var require:any;
 const cheerio = require('cheerio') as CheerioAPI;
 
+export function fetchAnnouncements() {
+  return (dispatch: Redux.Dispatch<any>): any => {
+    $.ajax({url: authSettings.urlBase + '/announcements',
```
smartin015: `Main.tsx` imports `actions/User`, which uses the `fetch` ambient function for web requests and has no problems. If we use that here, would it fix tests?

If that doesn't work, I recommend we either move this somewhere that doesn't break tests, or disable Main.test.tsx, adding a bug and a todo to ensure we fix our unit tests after Gencon. I'm really not a fan of every PR from now on showing broken tests, since we're inevitably going to miss test failures that way.

app/Style.scss
```diff
@@ -693,6 +693,28 @@ body {
         font-size: fontsize(flavor) !important;
       }
     }
+
+    &.announcing .logo {
+      // 2 lines of text + small top + bottom padding for button + top margin for logo
+      top: 2 * $lineheight * fontsize(interactive) + 3 * size(base);
```
smartin015: Recommend comment be `(2 text lines) + (small top & bottom padding and top margin for logo)`... at least that's how I read the code, anyways, not sure if correct.

app/reducers/Announcement.test.tsx
```diff
@@ -0,0 +1,3 @@
+describe('Snackbar reducer', () => {
```
smartin015: "Announcement pane"

app/actions/Web.tsx
toddmedema: Switching to fetch didn't work - I think we'd have to switch the rest of the functions in /web to also use fetch; so, moved it to /actions/Announcement, which is just as reasonable a place for it :)

app/Style.scss
toddmedema: Yep, definitely clearer (and correct)

PR #416, opened 2017-08-23
Fix several performance issues in rendering code

smartin015:
Before (during timer):

runaway heap growth example pattern

After (long period of interaction, followed by GC):

corrected heap growth with GC at end

Notice that the final result memory is roughly the same as before the long interaction with combat and cards.


toddmedema:
Looks good to me! While testing this, came across a bug, but seems unrelated to this / happens on prod too - https://github.com/ExpeditionRPG/expedition-app/issues/417

Issue #412, opened 2017-08-22
Show more detail on community quests in-app


smartin015:
In more quests section, mention that the quest ecosystem is community-driven and add a link to mobile-friendly details on the quest creator.

PR #419, opened 2017-08-24
improve community quest text

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/412


app/components/FeaturedQuests.tsx
```diff
@@ -36,7 +36,7 @@ const FeaturedQuests = (props: FeaturedQuestsProps): JSX.Element => {
       <Button onTouchTap={() => props.onSearchSelect(props.user, props.players)}>
         <div className="advanced_play">
```
smartin015: Can we rename the advanced_play class here?

toddmedema: Yep! Updated to questButton and questButtonWithIcon for clarity

Issue #405, opened 2017-08-17
Accessibility/Personalization: font size


smartin015:
Add text/ui size setting option. Most pages are slightly over length of iPad 2, require scrolling

Issue #124, opened 2016-09-08
Variable round speeds


smartin015:
Notify slayerofable@gmail.com when "lightning rounds/strategic play" game modes are available

toddmedema:
Following up here (based on feedback from backer) that it'd make sense to split out round speed from difficulty, ie keep difficulty but have it not affect timer duration, then make "Round speed" with options "strategic (60s), slow (30s), classic (10s), lightning (6s)"

smartin015:
Backlogging this for now.

PR #423, opened 2017-08-25
add timer speed and font size settings

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/405
Closes https://github.com/ExpeditionRPG/expedition-app/issues/124

screen shot 2017-08-25 at 12 26 50 pm
screen shot 2017-08-25 at 12 27 04 pm
screen shot 2017-08-25 at 12 27 14 pm
screen shot 2017-08-25 at 12 26 59 pm


app/components/Settings.tsx
```diff
@@ -10,31 +9,47 @@ export interface SettingsStateProps extends SettingsType {}
...
 
+export const fontSizeValues: FontSizeType[] = ['SMALL', 'NORMAL', 'LARGE'];
+
+const timerText: { [v: string]: any } = [
+  {title: 'Disabled', text: 'Timer disabled.' },
+  {title: 'Relaxed', text: 'Take your time, you have 30 seconds.' },
+  {title: 'Normal', text: 'Classic Expedition, 10 seconds per round.' },
+  {title: 'Fast', text: 'Act fast! 6 seconds per round.' },
+];
+export const timerValues: number[] = [999, 30, 10, 6];
```
smartin015: I'd prefer `null` or similar for a disabled timer, instead of a signal value.

app/components/Settings.tsx
```diff
@@ -10,31 +9,47 @@ export interface SettingsStateProps extends SettingsType {}
...
+
 const Settings = (props: SettingsProps): JSX.Element => {
   const difficultyIdx = ['EASY', 'NORMAL', 'HARD', 'IMPOSSIBLE'].indexOf(props.difficulty);
+  const fontSizeIdx = fontSizeValues.indexOf(props.fontSize);
+  const timerIdx = timerValues.indexOf(props.timerSeconds);
 
   const multitouchText = (props.multitouch) ? 'All players must hold their finger on the screen to end combat.' : 'A single tap will end combat.';
   const helpText = (props.showHelp) ? 'Setup and combat hints are shown.' : 'Setup and combat hints are hidden.';
+  const playerText = (props.numPlayers > 1) ? 'The number of players.' : 'Solo play: Play as two adventurers, and get twice as long in combat.';
```
smartin015: `twice as long to play`?

app/components/Settings.tsx
toddmedema: Good call. Updated to also skip the timer card completely if disabled

app/cardtemplates/combat/Combat.tsx
```diff
@@ -101,12 +104,45 @@ function renderDrawEnemies(props: CombatProps): JSX.Element {
...
+  if (props.settings.showHelp) {
+    helpText = (
+      <ol>
+        <li>Shuffle your ability draw pile.
+          <ul>
+            <li>Keep abilities played this combat in a separate discard pile.</li>
+            <li>If you run out of abilities to draw, shuffle in your discard pile.</li>
+          </ul>
+        </li>
+        <li>No timer: Draw three abilities from your draw pile and play one ability.</li>
```
smartin015: Probably don't need to say "No timer" here, since the user specifically selected this combat play style.

toddmedema: Was going along the lines of highlighting rules that are specific to 1 player / 2 player, which is there to indicate that this rule is different than usual, so that if players do play on a different mode, they might expect it to be different

Issue #424, opened 2017-08-26
Show retry button if combat -> one RP card -> end


toddmedema:
in roleplay/Roleplay.tsx

Likely easiest by passing the previous ParserNode to loadRoleplayNode, checking if that's combat, then checking if the next node isEnd from parser/Handlers.tsx

Issue #398, opened 2017-08-04
Combat defeat page upgrades


toddmedema:
If it's the last combat / next link is end, add a retry button

Helpful reminders / suggestions: (perhaps show one random each time?)
- Healing can be done on adventurers at 0 health.
- Look for combinations in your abilities - two adventurers working together can do more damage than two alone.
- Battles are never won by healing - only by damaging the enemy.

PR #425, opened 2017-08-26
Add retry on immediate-end combat loses and show tips on defeat page

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/398
Closes https://github.com/ExpeditionRPG/expedition-app/issues/424


app/cardtemplates/combat/Combat.tsx
```diff
@@ -301,15 +303,34 @@ function renderVictory(props: CombatProps): JSX.Element {
...
+  // Use a random number in the state to keep it consistent / not change on new render events
+  const helpText = helpfulHints[props.mostRecentRolls[0] % helpfulHints.length];
+
+  // If onLose is just an end, offer a retry button
+  // TODO: Figure out how to detect single rp card -> end, and also show a retry on that
+  // https://github.com/ExpeditionRPG/expedition-app/issues/424
+  let retryButton = <span></span>;
+  if (!props.custom) {
+    const nextNode = handleAction(props.node, 'lose');
+    if ((nextNode.getTag() === 'trigger' && nextNode.getNext() === null) || (nextNode.getNext() !== null && nextNode.getNext().getNext() === null)) {
```
smartin015: I'd recommend breaking out the end-trigger check from [Quest.tsx](https://github.com/ExpeditionRPG/expedition-app/blob/03ba7ca04742a1c00824f9d22e7540bc5ade3042/app/actions/Quest.tsx#L30), putting it in [app/parser/Handlers.tsx](https://github.com/ExpeditionRPG/expedition-app/blob/03ba7ca04742a1c00824f9d22e7540bc5ade3042/app/parser/Handlers.tsx) as a helper, and calling it here. That way, we have a standard way of detecting end nodes.

I think retries on the combat -> rp -> end case should be shown on the roleplaying card, so they aren't just wasted efforts by the quest writers. Can you move the todo to [loadRoleplayNode](https://github.com/ExpeditionRPG/expedition-app/blob/03ba7ca04742a1c00824f9d22e7540bc5ade3042/app/cardtemplates/roleplay/Roleplay.tsx#L37)?

It'd probably be easiest to implement by passing in the previous `ParserNode` (retrieved from state._history) to `loadRoleplayNode`, since it's non-trivial to get a previous node from a `ParserNode`.

toddmedema: @smartin015 good idea! Applied those updates, and was also able to implement for single RP cards, check it ;D

app/cardtemplates/roleplay/Roleplay.tsx
```diff
@@ -141,6 +144,15 @@ const Roleplay = (props: RoleplayProps, theme: CardThemeType = 'LIGHT'): JSX.Ele
     );
   });
 
+  // If they just got out of combat and the quest is about to end, offer the choice to retry combat
```
smartin015: s/they/we - pretty sure this tense matches comments elsewhere

app/cardtemplates/roleplay/Roleplay.tsx
```diff
@@ -141,6 +144,15 @@ const Roleplay = (props: RoleplayProps, theme: CardThemeType = 'LIGHT'): JSX.Ele
...
+  if (props.prevNode && props.prevNode.getTag() === 'combat' && rpResult.choices[0] && rpResult.choices[0].text === 'The End') {
```
smartin015: Casing against 'The End' here seems fragile (e.g. a manually crafted choice with the same text could cause this button to show). Could you instead check that the only next choice is an end trigger node using `isEnd()`?

app/parser/Handlers.tsx
```diff
@@ -72,3 +72,8 @@ export function handleAction(pnode: ParserNode, action: number|string): ParserNo
   }
   return pnode;
 }
+
+// Returns if the supplied node is an **end** trigger
+export function isEnd(pnode: ParserNode): Boolean {
```
smartin015: If you don't feel strongly about it... I'd prefer `isEndNode`

app/cardtemplates/roleplay/Roleplay.tsx
```diff
@@ -141,6 +144,15 @@ const Roleplay = (props: RoleplayProps, theme: CardThemeType = 'LIGHT'): JSX.Ele
     );
   });
 
+  // If we just got out of combat and the quest is about to end, offer the choice to retry combat
+  if (props.prevNode && props.prevNode.getTag() === 'combat' && isEndNode(props.node.getNext())) {
```
smartin015: Also need a check on the number of possible next choices here as well (i.e. only try looking for an end trigger if there's only a single path from this node, otherwise the user has branching choices to make and we don't need to show the retry).

PR #431, opened 2017-08-31
Added private search button in Tools

smartin015:
Included in this CL:
- Made all options in SearchSettings optional
- Extended the existing search results page to optionally show/hide the filters in the header, then used it in a new button on the Tools page to search specifically for Private quests.
- This'll require a change to the API server to do what's expected when passed "owner" as a search parameter. Coming soon.


app/components/Tools.tsx
```diff
@@ -29,6 +31,12 @@ const Tools = (props: ToolsProps): JSX.Element => {
           <div className="summary">Write your own quests and share them with the world.</div>
         </div>
       </Button>
+      <Button onTouchTap={() => props.onPrivateQuestsSelect(props.user)}>
+        <div className="advanced_play">
+          <div className="title">Privately Published Quests</div>
```
toddmedema: Just checking since this seems like a long title, but it fits on one line, even on an iPhone 5? (in the browser device emulator)

app/reducers/Search.tsx
```diff
@@ -18,6 +18,12 @@ export const initial_state: SearchState = {
 
 export function search(state: SearchState = initial_state, action: Redux.Action): SearchState {
   switch(action.type) {
+    case 'NAVIGATE':
+      // When viewing private quests, clear the searched quests first.
+      const nav = (action as NavigateAction);
+      if (nav.to && nav.to.name === 'SEARCH_CARD' && nav.to.phase === 'PRIVATE') {
```
toddmedema: Is there a reason that we don't always clear search results, instead of just when private searching? (For example, the inverse of going PRIVATE -> PUBLIC search)

app/components/Tools.tsx
smartin015: Yep, it's too long. Updated.

app/reducers/Search.tsx
smartin015: Good point - made the clearing of old results more explicit, and it's now done whenever a new search is started.

smartin015:
Ready for review again :)

app/components/Tools.test.tsx
```diff
@@ -1,15 +1,15 @@
 import * as React from 'react'
 import {shallow} from 'enzyme'
 import Tools, {ToolsProps} from './Tools'
-import {initial_state as settings_initial_state} from '../reducers/Settings'
-import {initial_state as user_initial_state} from '../reducers/User'
+import {initialUser} from '../reducers/User'
```
toddmedema: Yes!

PR #442, opened 2017-09-09
Add expedition-qdl repo, remove all QDL parsing in this repo

smartin015:


smartin015:
Merging this now, since it's basically a combo of 1300-line cleanup and no-op. Tests pass, confirmed working with all offline quests & an online quest.

Issue #426, opened 2017-08-27
Turn "Are you sure you want to exit the quest?" into themed modal (also fixes double confirm bug)


toddmedema:
Bug: asks if you want to leave twice when exiting a quest via the menu -> Home

toddmedema:
Turns out window.confirm results in a second click being fired (or a click end?).

We could do a weird, time consuming and janky hack....

OR, just turn this into a themed modal for about the same work + a much better UX / UI

toddmedema:
Engineer in such a way that we can support a variety of dialog content (such as reporting a quest, paying a tip, etc)

PR #445, opened 2017-09-09
Replace window.confirm with dialog

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/426

Architected such that we'll be able to put in arbitrary contents and actions in the future (such as for payments)

Example:

screen shot 2017-09-09 at 6 08 50 pm


app/actions/Dialog.tsx
```diff
@@ -0,0 +1,11 @@
+import {DialogCloseAction, DialogOpenAction} from './ActionTypes'
+
+export function closeDialog(): DialogCloseAction {
+  return {type: 'DIALOG_CLOSE'};
+}
+
+export function openDialog(title: string, content: JSX.Element, actions: JSX.Element[]): DialogOpenAction {
+
+console.log(title, content, actions);
```
smartin015: rm

app/components/Search.tsx
```diff
@@ -232,7 +232,7 @@ function renderResults(props: SearchProps, hideHeader?: boolean): JSX.Element {
       title="Quest Search Results"
       header={!hideHeader && <div className="searchHeader">
         <span>{props.results.length} quests for {props.numPlayers} <img className="inline_icon" src="images/adventurer_small.svg"/></span>
-        <Button className="float_right filter_button" onTouchTap={() => props.onFilter()}>Filter ></Button>
+        <Button className="floatRight filter_button" onTouchTap={() => props.onFilter()}>Filter ></Button>
```
smartin015: all other class names in this snippet are undercase... what's our style here?

app/components/base/Card.tsx
```diff
@@ -38,19 +40,26 @@ export default class ExpeditionCard extends React.Component<ExpeditionCardProps,
...
-        let result = true;
-        if (this.props.inQuest) {
-          result = window.confirm('Are you sure you want to exit this quest?');
+        if (!this.props.inQuest) {
+          return dispatch(toPrevious('SPLASH_CARD', undefined, false));
+        } else {
+          return dispatch(openDialog('Exit quest?',
+            <p>Tapping exit will lose your place in the quest and return you to the home screen.</p>,
+            [
+              <FlatButton onTouchTap={()=>dispatch(closeDialog())}>Cancel</FlatButton>,
```
smartin015: This seems like a lot of JSX to be storing in the reducer... typically the rendering is done by components, and we'd have something like e.g. [Dialogs.tsx](https://github.com/ExpeditionRPG/expedition-quest-creator/blob/master/app/components/Dialogs.tsx) so that it's relatively easy to find what's actually rendering content to the page. Can we use that method instead?

app/components/Search.tsx
toddmedema: Hmm. Think I've added a few camelCase, though apparently people these days think you shouldn't do that in CSS.

In other news, floatRight is only used in one place, and floatLeft isn't used anywhere, so was able to get rid of both classes by just properly assigning a position to .filter_button

app/components/base/Card.tsx
toddmedema: Updated! Incidentally, this ended up removing a bunch of code elsewhere in the app - now the main container doesn't have to know anything about the dialog, so I was able to get rid of all that weird jankiness about having to setTimeout after closing the dialog :)

toddmedema:
@smartin015 tests fixed, ready for re-review

Issue #448, opened 2017-09-10
Quest details page: show more details!


toddmedema:
At the very least, review starts, duration, genre and age rating - aka the main metadata

PR #449, opened 2017-09-10
improved search details page

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/448

screen shot 2017-09-10 at 10 28 36 am


toddmedema:
I decided to put the play button above the details since we'll be adding more content there as we go (such as the actual reviews people wrote, and it's kind of like an add to cart button where we want it to be as easy as possible for people to click it :)

Issue #404, opened 2017-08-17
Splash player count indicator


smartin015:
Indicate count of fingers/players on splash page (as people press, change the number)

toddmedema:
I was thinking that this would be specifically for multiplayer, and it'd be on the page after the multi-touch part - since it's hard to see a number on the screen when your hand + finger is covering most of it

smartin015:
As (I think) we talked about during the convention, having some part of the UI change when new fingers are added/removed and indication of when the "next card" action is about to trigger would improve the UX a bit here.

toddmedema:
Note: this doesn't make as much sense for combat, since the action triggers as soon as you have the appropriate number of fingers on the page, and there's already a number (timer) in the middle of the screen.

So, this will be just for the splash screen:
- show # of fingers on screen in the middle
- surround it with a big loading circle that fills in / spins to completion to indicate how much time is left before the page will change

toddmedema:
Also, remove "no multi-touch" button and replace with instructional text "Double tap / click to set number of players manually"

Then, add gesture awareness such that double tap / clicking goes to player set page. Maybe also mention on that page "Manually setting party size defaults to disabling multi-touch mode. You can enable multi-touch at any time by going to settings."

PR #452, opened 2017-09-10
improved splash touch experience

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/404

It's also on beta.expedition if you want to try it out :)

screen shot 2017-09-10 at 1 07 39 pm


toddmedema:
Woot! Not too worried about that, since it only happens when there's at least two fingers on the screen, so you probably can't see it anyway

Issue #444, opened 2017-09-09
Web app: handle back button


toddmedema:
https://stackoverflow.com/questions/136937/is-there-a-way-to-catch-the-back-button-event-in-javascript

Wire that into the existing toPrevious handling; can't use window.pushState since that wouldn't record / restore the whole state.

PR #453, opened 2017-09-10
treat browser back button as app back button

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/444


app/Main.tsx
```diff
@@ -173,10 +173,15 @@ function setupGoogleAnalytics() {
 }
 
 export function init() {
-  // TODO: remove these and have everyone use getPlatform()/getVersion().
+  // Setup as web platofrm as default; we might find out later we're an app
```
smartin015: Nit: `platofrm`

app/reducers/Card.tsx
```diff
@@ -3,12 +3,18 @@ import {CardState} from './StateTypes'
...
+
 // ts: 0 solves an obscure bug (instead of Date.now()) where rapidly triggering navigations with undefined states
 // (specifically from the editor) wouldn't work b/c their ts diffs were < DEBOUNCE
 export function card(state: CardState = {name: 'SPLASH_CARD', ts: 0}, action: Redux.Action): CardState {
   switch (action.type) {
     case 'NAVIGATE':
-      const to = (action as NavigateAction).to;
+      const navigateAction = (action as NavigateAction);
+      if (historyApi) {
+        history.pushState(null, '', '#' + navigateAction.to.name);
```
smartin015: This'll probably break for instances of moving between phases of a single  card (e.g. combat timer to resolve card) since the  name is the same. Suggest using the action timestamp instead.

Issue #428, opened 2017-08-28
Allow quests to skip the intro/setup card


smartin015:
Future quests (such as the one by Louise and Earl) will have specific instructions on how to set up the game - drawing specific adventurers, abilities, loot, and/or persona cards. Without skipping the initial instruction card, adventurers would end up doing unnecessary extra setup work.

One way to do this would be to have something like `<quest skipSetup="true">` which just spits the user out into the main storyline. The quest creator can add a few initial cards of setup at their discretion.

toddmedema:
Since the QC works in markdown instead of XML, perhaps this returns to that previous metadata implementation we had, where it's fields below the title?
```
_Title of quest_
skipSetup: true
```

Or even just for boolean flags that default to no:
```
_Title of quest_
skipSetup
```

Something to consider here is how we'd want to do other types of metadata in the future, such as (listing required media files?)... though, looking through our backlog, I can't really find any other metadata things to consider...


Another idea: What if it was an ops function, and we added the ability to call ops functions as soon as the quest started / before even the setup card was shown?

smartin015:
Perhaps. I'm more amenable to just using JSON for now, since this will probably not be a common feature and that sort of metadata implementation will take significantly more work.

toddmedema:
So, `_Title of quest_ {skipSetup: true}`? That seems reasonable, since it's a pretty advanced feature

PR #468, opened 2017-09-14
add ability for quests to have custom _Setup_ card at start

toddmedema:
@smartin015 you'll like this one ;D

Closes https://github.com/ExpeditionRPG/expedition-app/issues/428


smartin015:
As discussed - let's move this from implicit intent (card with "setup" title) to explicit intent (A `{"skipSetup": true}` or similar JSON field after the card).

PR #471, opened 2017-09-15
move clone to fix round offset bug

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-qdl/issues/13#issuecomment-329765986

Given your input, this seems like an even simpler solution - just moving the clone to after roundCount is incremented. Also solves the valid concern about double-running ops that would happen if we attempted to run `.clone` twice


app/cardtemplates/combat/Actions.tsx
```diff
@@ -271,10 +272,10 @@ export function midCombatChoice(settings: SettingsType, parent: ParserNode, inde
 
 export function handleCombatTimerStop(node: ParserNode, settings: SettingsType, elapsedMillis: number) {
   return (dispatch: Redux.Dispatch<any>): any => {
-    node = node.clone();
     node.ctx.templates.combat.mostRecentAttack = generateCombatAttack(node, settings, elapsedMillis);
     node.ctx.templates.combat.mostRecentRolls = generateRolls(settings.numPlayers);
     node.ctx.templates.combat.roundCount++;
+    node = node.clone();
```
smartin015: This increments the round and changes attack/rolls for the previous card, which breaks partitioning of nodes across the history boundary. Really, cloning should be the first action before any mutation is made to an existing parser node. 

I'd much prefer we move the clone call as indicated in `handleResolvePhase` for now.

toddmedema:
Detailing my thoughts for discussion tonight:

- This PR adds one clone, done immediately after the node ctx is adjusted: https://github.com/ExpeditionRPG/expedition-app/pull/471/files
- Which is similar to your suggestion about moving the `clone()` to be before the check to visibleKeys(https://github.com/ExpeditionRPG/expedition-qdl/issues/13#issuecomment-329765986), except that it clones the node immediately after it is changed, instead of before a function is run on the node that should have no impact on it
- If you look at the order of operations, it's pretty much identical (it's either `roundCount++` -> clone -> handleResolvePhase, or `roundCount++` -> handleResolvePhase -> clone), but I believe the former to be clearer because it's showing "Ok, we modified the node, now we need to re-clone it", whereas the latter is showing "Well, we haven't modified the node... and we're about to run a function that shouldn't change anything about the node.... but we should clone it just in case?"
- TL;DR I think that cloning the node immediately after a change seems to better indicate what we're doing with the clone (re-calculating internals after ctx is changed)
- If we wanted to take a bit longer to do this in a more principled manner, perhaps we might replace all edits to node.ctx with a node.updateContext function that would then re-calculate any internals that rely on ctx? (Perhaps you're able to pass it a dictionary of changes, so that the calculations could happen all at once)
  - That being said, that wouldn't solve the `random()` problem.... not really sure how to solve that. For what it's worth, `.random()` seems like it should be totally broken right now given how many calls to `.clone()` there are in combat/Actions, and that every one of those would result in a re-evaluation of ops

smartin015:
Here are my thoughts on how nodes should behave (not necessarily how they do right now):

1. All op evaluations done on the transition between cards should be done on the first card's rendered node state and context (i.e. we don't update the round, *then* check on visiblekeys).
2. A node should not mutate after it's saved into the redux store (i.e. mid-card affects should not affect the current stored node or history nodes)
3. A node should only be cloned once (when mutation starts), and all mutation should be done before it's inserted into the redux store.

This can be maintained at compile time by having two class types, `ParserNode` and `MutableParserNode`. 

- The `ParserNode` marks its properties [readonly](https://basarat.gitbooks.io/typescript/docs/types/readonly.html), pre-renders at construction time, and has a `clone()` method that returns a `MutableParserNode`.
- The latter has mutable properties, uses its immutable parent's rendered node set (does not re-render), and has no `clone()` method (i.e. can't clone more than once). It also has a `finalize()` method that returns an immutable `ParserNode`.

Typical flow would be: action dispatched -> action calls `node.clone()` -> mutates node -> calls `mnode.finalize()` & returns the constructed action to the reducer.

PR #475, opened 2017-09-17
Remote Play Demo

smartin015:


smartin015:
Status update: 

- super-basic session system created (allows connection via secret code)
- basic touch point communication done (handles ontouchstart)

video of progress: [LINK](https://drive.google.com/open?id=0BzrQOdaJcH9MbWF2S0lWOFl0Qm8)

app/actions/Web.tsx
```diff
@@ -148,3 +150,98 @@ export function submitUserFeedback(quest: QuestState, settings: SettingsType, us
...
+}
+
+export function remotePlayNewSession(user: UserState) {
+  return (dispatch: Redux.Dispatch<any>): any => {
+    fetch(remotePlaySettings.newSessionURI, {
+      method: 'POST',
+      mode: 'cors',
+      headers: new Headers({
+        'Accept': 'text/html',
+        'Content-Type': 'text/html',
```
toddmedema: Content type is html, but it parses it as JSON on line 172?

app/components/RemotePlay.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const history = this.props.remotePlay.history.map((m: SessionMetadata, i: number) => {
+      return (<Button key={i} onTouchTap={()=>{this.props.onReconnect(this.props.user, m.id)}}>
+        {m.questTitle} ({m.peerCount} peers) - {m.firstContact}</Button>);
+    });
+
+    return (
+      <Card title="Remote Play">
+        <div className="remoteplay">
+          <div className="helptext">Lorem Ipsum Connect Here:</div>
+          <input type="text" autoComplete="off" autoCorrect="off" autoCapitalize="off" spellCheck={false} onChange={(e: any) => {this.handleSecret(e)}} value={this.state.secret}></input>
```
toddmedema: Would it be worth making these codes limited to numbers? Would avoid confusion around 1 vs l and 0 vs O.

Nevermind, looks like we wouldn't be able to generate enough entropy. Perhaps in the generation code we can just remove 0 / 1 / I / L from the character options.

app/components/RemotePlayContainer.tsx
```diff
@@ -0,0 +1,45 @@
...
+    phase: ownProps.phase,
+    user: state.user,
+    remotePlay: state.remotePlay,
+  };
+}
+
+const mapDispatchToProps = (dispatch: Redux.Dispatch<any>, ownProps: any): RemotePlayDispatchProps => {
+  return {
+    onConnect: (user: UserState, secret: SessionSecret) => {
+      if (secret.length !== 4) {
```
toddmedema: If secret length is 4 alphanumeric characters (excluding 0/o/i/1), that's 32^4, only 1M possible options. With just 200 items, the chance of a collision is nearly 2%. And there'll be almost 10,000 copies in play by this Christmas.

I would be in favor of increasing length to 6, which would be able to handle over 6,000 simultaneous sessions with <2% collision chance.

app/components/RemotePlay.tsx
smartin015: They're actually currently limited to capital letters A-Z, for `26^4 = 456k` possible combinations.

Since we're locking down the session, each session would only be receptive to attack for about 5-10 minutes or so. 456k is plenty, especially if we rate-limit (and we eventually should, but this is a cobbled together demo so far)

app/components/RemotePlayContainer.tsx
smartin015: In the final version, I was toying with a different approach than random: using `http://hashids.org/` plus a shuffled sequence of `1...456976` and a counter that iterates over the sequence and applies the hash. We can randomize the shuffle each time to further reduce the likelihood of a rollover collision.

Under normal operation, we'd have to have 456k active, unlocked sessions in order to have a collision of secrets. Which seems pretty good to me, for now.

toddmedema: I like this plan - since sessions will only be joinable for 10 minutes or less, that solves the security concern - then it's just down to properly handling / preventing duplicate IDs from being active at the same time.

smartin015:
I'm still working on fixing tests, but the overall code is ready for review.

Keep in mind this is currently dev-only, and doesn't fully implement remote play (just uses the new format for card nav, user login, and combat end actions). Another PR will be needed to migrate actions to the new remote play format, plus another to flesh out the remote play UI, plus another for additional multitouch propagation, etc etc etc.

Issue #386, opened 2017-07-15
Expansion UX


toddmedema:
Once app PR live:
- QC: update app version, make sure {{_.contentSets()}} works (though keep hidden for now)
- QC: if quest set to require "The Horror", enable it in app settings via renderAndPlay

At the start, ask if they have expansion (if setting.contentSets || contentSets.horror undefined)
- DONE: Store as dictionary of enabled content, ie {horror: null / true / false} - use null state to tell if they've been asked yet or not
- DONE: Save to settings & localStorage
- DONE: Modal. Noting that this can be changed at any time in settings
- DONE: Ability to change in settings

If true:
- DONE: Web: showcommunity quests that are expansion-pack-only
- DONE: Add a new "Learning to Horror" quest that specifically teaches the new Horror / Persona mechanics
  - Update featured quests page to only show it if horror enabled
  - Each persona card has 5 levels; the intermediate levels (indicated by arrows) offer no bonuses. You cannot go over Max or under Min. Right before you resolve abilities, any players at Min / Max will resolve their effect and then return to Base persona.
  - Your persona reveals details about your character(s) and how they react to events and situations in the game. As your persona changes, your character(s) will react differently to the world around them. For players who enjoy roleplaying, this adds a lot of new depth to characters. Of course, roleplaying is not required.
  - Influence abilities and Horror surges will raise and lower your persona level. Your decisions in quests can also affect your persona - if they do, the quest will tell you.
  - Persona levels persist across fights.
- DONE: review / update expansion rules (https://docs.google.com/document/d/1n-zMfg9mHNFAY6LSgt-IygzipwfBNd2y576Qqumtfr8/edit) and apply them to the app
  - Quest Setup: draw a persona and set to Base. Everyone draw 1 influence ability (start with 7 abilities).
  - resolve phase: resolve any personas at min / max, then return them to base (before resolving abilities)
  - adventurer phase: if alive players < numPlayers, show current hp = 0 rules, plus resetting persona
  - end of combat leveling up: all adventurers can also learn influence abilities.

Consider:
- DONE: You may choose when to spend max persona? Ie doesn't have to be the very next round. more strategy around use + a bit of a gambling / push your luck mechanic

Behind the scenes:
- DONE: add new encounters to config list
- DONE: make sure they only show up in .randomEnemy() if horror is enabled
- DONE: Quest Creator: new button and DB column for requires_the_horror

toddmedema:
[...]

PR #477, opened 2017-09-19
Add Expansion UX, Rules, Setting, Learning Quest

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/386

On top of adding all of the new expansion UI things, this also fixes some small annoying UI bugs, primarily that picker titles had their line height set really large to vertically center the text, but that broke when the text became two lines. Looks much better now! Also improves the display of dialogs (adding the papyrus background + better button styling), and the combat damage page (highlights damage, and doesn't show the "knocked out?" help text unless alive adventurers < players).

screen shot 2017-09-19 at 2 42 21 pm
screen shot 2017-09-19 at 2 42 45 pm
screen shot 2017-09-19 at 2 23 14 pm
screen shot 2017-09-19 at 2 15 39 pm
screen shot 2017-09-19 at 2 14 27 pm
screen shot 2017-09-19 at 2 41 43 pm
screen shot 2017-09-19 at 2 05 09 pm
screen shot 2017-09-19 at 2 04 54 pm


toddmedema:
Improved prompt subtext: 
screen shot 2017-09-21 at 8 19 30 pm

app/Globals.test.tsx
```diff
@@ -29,4 +29,17 @@ describe('Globals', () => {
...
   });
+
+  describe('getStorageKey', () => {
+    it('Boolean');
+    it('JSON');
+    it('Number');
+    it('String');
+  });
+  describe('setStorageKey', () => {
+    it('Converts booleans to strings');
```
smartin015: No caps on start of it() string

app/Globals.tsx
```diff
@@ -91,9 +91,15 @@ export function setNavigator(navigator: any) {
   refs.navigator = navigator;
 }
 
+// accepts boolean, numbers, strings and JSON containing only those types
```
smartin015: Comment doesn't match the function implementation

app/Globals.tsx
```diff
@@ -91,9 +91,15 @@ export function setNavigator(navigator: any) {
...
 export function setStorageKey(key: string, value: any) {
```
smartin015: `setStorageItem` maybe? This isn't just setting a key.

app/Globals.tsx
```diff
@@ -91,9 +91,15 @@ export function setNavigator(navigator: any) {
...
   try {
-    getWindow().localStorage.setItem(key, value.toString());
+    if (typeof value === 'object') {
+      value = JSON.stringify(value);
+    } else {
+      value = value.toString();
+    }
+    getWindow().localStorage.setItem(key, value);
   } catch (e) {
     console.log('Error setting storage key', key, 'to', value, e);
```
smartin015: `console.error`

app/Globals.tsx
```diff
@@ -129,6 +135,11 @@ export function getStorageKeyBoolean(key: string, fallback: boolean): boolean {
   return (val !== null) ? (val.toLowerCase() === 'true') : fallback;
 }
 
+export function getStorageKeyJson(key: string, fallback?: object): object {
```
smartin015: Same here... the returned values aren't keys

app/Main.tsx
```diff
@@ -195,6 +196,14 @@ export function init() {
   setupGoogleAnalytics();
   getStore().dispatch(fetchAnnouncements());
 
+  const contentSets = (getStore().getState().settings || {}).contentSets;
+  for (const set in contentSets) {
```
smartin015: So if settings fails to load, the for loop and modal are skipped? Is this the behavior we want?

app/cardtemplates/combat/Combat.tsx
```diff
@@ -211,9 +211,11 @@ function renderSurge(props: CombatProps): JSX.Element {
 
 function renderResolve(props: CombatProps): JSX.Element {
   let helpText: JSX.Element = (<p>Resolve all played abilities.</p>);
+  const theHorror = (props.settings.contentSets.horror === true);
   if (props.settings.showHelp) {
     helpText = (
       <span>
+        {theHorror && <Callout icon="horror_white"><strong>The Horror:</strong> Any adventurers at Min persona must resolve their persona effect and then return to Base before resolving abilities. Adventurers at Max persona may choose to resolve and reset now or in a later round.</Callout>}
```
smartin015: `to Base persona`

app/cardtemplates/combat/Combat.tsx
```diff
@@ -296,6 +300,7 @@ function renderPlayerTier(props: CombatProps): JSX.Element {
 
 function renderVictory(props: CombatProps): JSX.Element {
   const contents: JSX.Element[] = [];
+  const theHorror = (props.settings.contentSets.horror === true);
```
smartin015: I'm seeing this repeated quite a bit. Is there a way we can generalize this? Export a `horrorExpansion()` boolean function somewhere?

app/cardtemplates/combat/State.test.tsx
```diff
@@ -2,19 +2,23 @@
 describe('Combat State', () => {
   describe('combatScope', () => {
     describe('randomEnemy', () => {
-      it('gets a random enemy');
+      it('gets a random enemy name');
```
smartin015: No implementation? :(

app/Globals.tsx
toddmedema: Hmm? That's what value can be. Updated the comment to "Value can be boolean, number, string or stringifiable JSON" to increase clarity

app/Globals.tsx
toddmedema: Updated to setStorageKeyValue since that seems to be the clearest description of what it's doing :)

app/Globals.tsx
toddmedema: Reduced to `getStorageJson`, etc

app/Main.tsx
toddmedema: Good point; updated to show the modal if settings fail to load.

app/cardtemplates/combat/Combat.tsx
toddmedema: Not sure that's worth it at this point since we'd be replacing a boolean check with a function call, and then have to worry about managing that function. Perhaps we can design a more generalizable solution in 6 months once we have multiple expansions and keeping track of them gets more complex :)

app/cardtemplates/combat/State.test.tsx
toddmedema: Ask and you shall receive :)

toddmedema:
@smartin015 all comments addressed!

app/Globals.tsx
smartin015: As implemented, it'll store anything that's an object type or has a toString function (i.e. it doesn't explicitly reject Array, or Set, etc). JSON is a string wire format, not a javascript object.

If you want to restrict to those types, update value param to be `number|boolean|Object` instead of `any`.

app/Main.tsx
```diff
@@ -196,12 +196,17 @@ export function init() {
...
+  if (settings) {
+    const contentSets = (settings || {}).contentSets;
+    for (const set in contentSets) {
+      if (contentSets[set] === null) {
+        getStore().dispatch(setDialog('EXPANSION_SELECT'));
+        break;
+      }
     }
+  } else {
+    getStore().dispatch(setDialog('EXPANSION_SELECT'));
```
smartin015: Can you add a test that asserts this dialog shows if settings is null and/or contentSets is weird?

Issue #467, opened 2017-09-13
V2 sfx & music engine


toddmedema:
PR 1:

- DONE: setting to disable / enable, including auto pause / resume
- DONE: convert AUDIO_SET action to accept a dictionary (like setSettings)
- DONE: control fading it in and out to prevent abrupt transitions
- DONE: know when one loop is done so that it can start the next loop
- DONE: smoothly play music even as you're sliding between cards, on devices that are several years old (it hiccups a bit on a 3 year old tablet, but not too bad)
- DONE: minimizing app / tab should fade out music; re-opening should resume

PR 2: better caching, navigation and prevent memory leaks:

// TODO: where to note that audio files should be encoded in Audacity with settings:
// Preset medium (145-185kbps), variable speed fast, 441000hz, force mono channel
// But we should definitely store the raw / HQ files somewhere.... (root/assets folder?)

- create a new caching action / reducer set; state.cache.audio for now (can be extended to quests and other assets as well)
- update audio actions and reducers to save to and pull from the cache
- like settings, make sure state.cache isn't duplicated and isn't overwritten when going forward and backward in history
- in combined reducers, when RETURNing, diff audio.loop contents (sorted) - if they're different, kick off the old loop
- Music Engine (aka in combat): hard-code filenames into logic
- Sound effects (triggered by inputs)

Sound assets advice from Nick:

- 2x 2 minute loops, one more intense (higher tier) and one less intense (lower tier), with a basic msuic engine that changes which loop is next based on combat tier
- 4 (~5s) riffs: surge, timer running out, victory, defeat
- 4 (~1s) sound effects:enemy added (trumpet?), enemy killed (crushing / attack sound), adventurer killed (wilhelm scream?), adventurer revived (heal sound)

toddmedema:
Inspirations:

- Existing Expedition music: https://www.kickstarter.com/projects/toddmedema/expedition-the-horror
- medium intensity, drums and woodwinds https://www.youtube.com/watch?v=1RBj-Yc4okY
- high intensity drums drums, (brass and strings?) https://www.youtube.com/watch?v=UXaYtKjSdz0&t=0s


Other thoughts:

- 60k stretch goal for additional, horror-specific music?

PR #478, opened 2017-09-22
Audio / music engine

toddmedema:
Addresses part 1 of https://github.com/ExpeditionRPG/expedition-app/issues/467 - which is to say, is the fundamental engine that keeps track of what's playing, starts, stops, etc.

Note that this is not meant to be deployed in its current state (see: test buttons still in code), and will require a second PR flushing out a few more mechanical details and adding in the app logic for when to play which files. But, that'll be about as big as this one, so figured it'd save your sanity to split them :)

DONE: API to play one-shot sound effects, as well as cue loops of music
DONE: setting to disable / enable, including auto pause / resume
DONE: fade music in and out to prevent abrupt transitions from settings changes, music ending, etc
DONE: allow looping music
DONE: performance: smoothly play music even as you're sliding between cards, on devices that are several years old (it hiccups a bit on a 3 year old tablet, but not too bad)
DONE: minimizing app / tab should fade out music; re-opening should resume


app/actions/Audio.tsx
```diff
@@ -0,0 +1,159 @@
...
+function audioSet(changes: any): AudioSetAction {
+  return {type: 'AUDIO_SET', changes} as AudioSetAction;
+}
+
+function audioStoreBuffer(name: string, buffer: any): AudioStoreBufferAction {
+  return {type: 'AUDIO_STORE_BUFFER', name, buffer};
+}
+
+function loadAudioFile(context: any, url: string, callback: (err: string, buffer: any) => void) {
+  const request = new XMLHttpRequest();
```
smartin015: Any reason why we're not using `fetch` here? Is this only going to be used for local files like with [fetching local quests](https://github.com/ExpeditionRPG/expedition-app/blob/a73326f1f45f009fbebdac69e896c27e7b5fb3f6/app/actions/Web.tsx#L22)?

Either way... suggest matching styles, and adding error handling on the request.

app/components/FeaturedQuests.tsx
```diff
@@ -36,6 +41,11 @@ const FeaturedQuests = (props: FeaturedQuestsProps): JSX.Element => {
 
   return (
     <Card title="Select Your Quest" icon="adventurer">
+
+<Button onTouchTap={() => props.musictest()}>Music</Button>
```
smartin015: Shall we gate this on `NODE_ENV=dev` and move it to the tools section? Otherwise, should probs remove.

app/components/SettingsContainer.tsx
```diff
@@ -26,6 +27,11 @@ const difficultySub: any = {
 
 const mapDispatchToProps = (dispatch: Redux.Dispatch<any>, ownProps: any): SettingsDispatchProps => {
   return {
+    onAudioChange: (v: boolean) => {
+      if (v) { dispatch(audioResume()); }
```
smartin015: Plz spread

app/reducers/Settings.tsx
```diff
@@ -4,6 +4,7 @@ import {ChangeSettingsAction} from '../actions/ActionTypes'
 import {getStorageBoolean, getStorageJson, getStorageNumber, getStorageString, setStorageKeyValue} from '../Globals'
 
 export const initialSettings: SettingsType = {
+  audioEnabled: getStorageBoolean('audioEnabled', true),
```
smartin015: I feel like we could make this a bit less redundant by declaring the object, then looping on keys (with values indicating the type & default) and calling the appropriate `getStorage*` method. Add a TODO?

app/reducers/StateTypes.tsx
```diff
@@ -10,6 +10,18 @@ export interface AnnouncementState {
   link: string;
 }
 
+export interface AudioState {
+  context: any;
+  currentMusic: any; // source node
```
smartin015: Why not `sourceNode`, `gainNode`, `loopFiles` etc? Prevents the need for descriptive comments and makes it easier to understand the code in other areas.

app/Globals.tsx
```diff
@@ -18,6 +18,8 @@ export interface ReactWindow extends Window {
   AndroidFullScreen?: {
     immersiveMode: (success: () => any, failure: () => any) => void,
   };
+  AudioContext?: any;
```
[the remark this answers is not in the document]
toddmedema: Unfortunately not, that's a variable that's created by the browser (old browsers) and we don't have control over the name

app/Main.tsx
```diff
@@ -208,8 +230,9 @@ export function init() {
   } else {
     getStore().dispatch(setDialog('EXPANSION_SELECT'));
   }
-
-  render();
+  getStore().dispatch(fetchAnnouncements());
+  const state = getStore().getState();
+  getStore().dispatch(audioLoad(state.audio, state.settings));
```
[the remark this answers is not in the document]
toddmedema: To get the current state.audio and state.settings. We could move the getStore().getState() into the audioLoad action if that would help?

app/actions/ActionTypes.tsx
```diff
@@ -10,6 +10,17 @@ export interface AnnouncementSetAction extends Redux.Action {
   link?: string;
 }
 
+export interface AudioSetAction extends Redux.Action {
+  type: 'AUDIO_SET';
+  changes: any;
+}
+
+export interface AudioStoreBufferAction extends Redux.Action {
```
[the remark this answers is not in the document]
toddmedema: Added comment next to the action function

app/actions/Audio.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+
+export function audioLoad(audio: AudioState, settings: SettingsType) {
+  return (dispatch: Redux.Dispatch<any>): any => {
+    try {
+      const context = new (getWindow().AudioContext || getWindow().webkitAudioContext)();
+      dispatch(audioSet({context}));
+
+      Object.keys(audio.buffers).forEach((file: string) => {
+        loadAudioFile(context, 'audio/' + file + '.mp3', (err: string, buffer: any) => {
```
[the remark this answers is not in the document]
toddmedema: Good call

app/actions/Audio.tsx
```diff
@@ -0,0 +1,159 @@
...
+import {logEvent} from '../Main'
+import {getStore} from '../Store'
+
+
+const MUSIC_FADE_SECONDS = 1.5;
+
+
+export function audioLoad(audio: AudioState, settings: SettingsType) {
+  return (dispatch: Redux.Dispatch<any>): any => {
+    try {
```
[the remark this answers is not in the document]
toddmedema: Length to complete on local dev with one music file and one SFX (on local dev): 800ms

Estimated length to complete with 15 music layers and 10 SFX (on local dev): 10 seconds

-> So, around 10 seconds with the app, potentially as long as 30+ seconds with the web app on a slow connection. Thankfully, it's loading in the background / not breaking anything. We could potentially add these files to a cache manifest to reduce repeat load as well.

My only concern with skipping audio loading now is that if they go to enable it later, it could be 30+ seconds before there's any indication it's working.

app/actions/Audio.tsx
```diff
@@ -0,0 +1,159 @@
...
+      }
+      logEvent('web_audio_api_err', err);
+      console.log('Web Audio API is not supported in this browser');
+    }
+  };
+}
+
+export function audioPause() {
+  return (dispatch: Redux.Dispatch<any>): any => {
+    const audio = getStore().getState().audio;
```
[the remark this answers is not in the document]
toddmedema: Alas, I do not know how to implement what you're discussing. The main reason I went this route was so that we don't have to make every other part of the app aware of the audio state.

app/actions/Audio.tsx
```diff
@@ -0,0 +1,159 @@
...
+    dispatch(audioPlayMusic(audio.loopPaused, audio));
+    dispatch(audioSet({loopPaused: []}));
+  }
+}
+
+// Immediately play the specified file
+export function audioPlayFx(file: string, audio: AudioState) {
+  return (dispatch: Redux.Dispatch<any>): any => {
+    if (!audio.context) { return console.log('Skipping playing audio ' + file + ', audio context failed to initialize'); }
```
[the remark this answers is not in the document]
toddmedema: Ah, was the oppressive whitespace tax repealed? Huzzah!

Done :)

app/actions/Audio.tsx
```diff
@@ -0,0 +1,159 @@
...
+  return (dispatch: Redux.Dispatch<any>): any => {
+    if (!getStore().getState().settings.audioEnabled) { return console.log('Skipping playing music, audio currently disabled'); }
+
+    const file = audio.loop.shift();
+    const buffer = audio.buffers[file];
+    audio.loop = audio.loop.concat(file);
+    dispatch(audioSet({loop: audio.loop}));
+
+    if (audio.currentMusicTimeout) {
+      clearTimeout(audio.currentMusicTimeout);
```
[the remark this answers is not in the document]
toddmedema: Right now, there's only ever a single music track playing. This will likely get upgraded with the new music layering concept, but that's going to end up as a whole PR of complexity on top of this architecture. We'll basically have to have / keep track of a timeout for every track that's currently playing, so that we can loop it.

app/actions/Audio.tsx
```diff
@@ -0,0 +1,159 @@
...
+// Plays the next song in the music loop with a cross-fade transition
+// If there's no songs in the loop, just fades out of the current song
+function audioPlayMusicLoopNext(audio: AudioState) {
+  return (dispatch: Redux.Dispatch<any>): any => {
+    if (!getStore().getState().settings.audioEnabled) { return console.log('Skipping playing music, audio currently disabled'); }
+
+    const file = audio.loop.shift();
+    const buffer = audio.buffers[file];
+    audio.loop = audio.loop.concat(file);
+    dispatch(audioSet({loop: audio.loop}));
```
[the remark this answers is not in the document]
toddmedema: Music loops will be 1 minute long, and only when music is playing. If that seems too much, let me know and we can discuss other implementation architectures

app/actions/Audio.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+    if (audio.currentMusicTimeout) {
+      clearTimeout(audio.currentMusicTimeout);
+    }
+
+    if (audio.currentMusicGain) {
```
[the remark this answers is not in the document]
toddmedema: Welcome to the joys of web audio :) Added a few more comments to attempt to clarify

app/actions/Audio.tsx
```diff
@@ -0,0 +1,159 @@
...
+    }, (buffer.duration - MUSIC_FADE_SECONDS) * 1000);
+    dispatch(audioSet({
+      currentMusic: audio.currentMusic,
+      currentMusicGain: audio.currentMusicGain,
+      currentMusicTimeout,
+    }));
+  }
+}
+
+function audioSet(changes: any): AudioSetAction {
```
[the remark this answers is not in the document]
toddmedema: Hmm. So it's based on the settings change action, where you can pass a dictionary containing any of the keys in the audio state object. I could make the AudioSetAction definition something like:
```
export interface AudioSetAction extends Redux.Action {
  type: 'AUDIO_SET';
  changes: {
      context?: any;
      currentMusic?: any; // source node
      currentMusicGain?: any; // gain node
      currentMusicTimeout?: any; // timeout
      loop?: string[]; // filenames
      loopPaused?: string[]; // filenames
      buffers?: {
        [key: string]: any; // filename: buffer
      };
  };
}
```
But my concern would be that it adds to the maintenance load by duplicating the audio state definition to yet another place.

app/actions/Audio.tsx
toddmedema: Same reason as fetch quests; added improvements from fetch quest to this function, and applied a few improvements from this function to fetch quest (and confirmed it still works). Also updated function name

app/reducers/Settings.tsx
toddmedema: Added

app/components/SettingsContainer.tsx
toddmedema: Done

app/components/FeaturedQuests.tsx
toddmedema: Good idea; perhaps as we add even more tests and public tools, that can eventually be split out to its own testing panel

app/reducers/StateTypes.tsx
toddmedema: Good idea, done

app/actions/Audio.tsx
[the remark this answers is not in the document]
smartin015: Why not `export interface AudioSetAction extends AudioState` or similar, with `AudioState` params marked as optional?

app/actions/Audio.tsx
```diff
@@ -86,50 +96,54 @@ export function audioPlayMusic(files: string[], audio: AudioState) {
...
+      clearTimeout(audio.musicTimeout);
     }
 
-    if (audio.currentMusicGain) {
-      audio.currentMusicGain.gain.linearRampToValueAtTime(1, audio.context.currentTime);
-      audio.currentMusicGain.gain.linearRampToValueAtTime(0, audio.context.currentTime + MUSIC_FADE_SECONDS);
-      audio.currentMusic.stop = audio.currentMusic.stop || audio.currentMusic.noteOff; // polyfill for old browsers
-      audio.currentMusic.stop(audio.context.currentTime + MUSIC_FADE_SECONDS);
+    if (audio.gainNode) {
+      audio.gainNode.gain.linearRampToValueAtTime(1, audio.context.currentTime); // Set the node gain to 1 now so that we can fade out soon
```
smartin015: `setValueAtTime`? (See below)

app/reducers/StateTypes.tsx
```diff
@@ -12,11 +12,11 @@ export interface AnnouncementState {
 
 export interface AudioState {
   context: any;
```
smartin015: Can we get actual types for these? https://www.npmjs.com/package/@types/webaudioapi

Also, if the audio context is part of the window, it doesn't need to be part of the redux state.

app/actions/Audio.tsx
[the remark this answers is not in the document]
smartin015: https://www.npmjs.com/package/react-sound and other libs treat audio as a react component (since it's technically part of the UI). What I meant by "invisible container" was to have a react component for the soundtrack that holds internal state including WebAudio nodes and position in the background music queue. The component renders an invisible element or span.

Then we can write `AudioContainer.tsx` or similar, which pulls details from redux store (like list of background music to play) and feeds it to the component.

This doesn't account for one-offs like SFX, but I think those can still be a simple action that sets up and runs the node independently of any background music components (since sound effects are independent and ephemeral).

Still leaves the matter of retrieving audio enabled/disabled from settings, but I think that's a far smaller (and already done) leak than the new audio context.

app/actions/Audio.tsx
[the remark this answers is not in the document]
toddmedema: Only loading from within the app if they're running the app - thinking of web app users

app/reducers/StateTypes.tsx
toddmedema: Added typing

Also, IMHO it does, because it needs to be instantiated: `context = new (getWindow().AudioContext as any || getWindow().webkitAudioContext as any)();`

app/actions/Audio.tsx
```diff
@@ -86,50 +96,54 @@ export function audioPlayMusic(files: string[], audio: AudioState) {
...
-    audio.currentMusic.start = audio.currentMusic.start || audio.currentMusic.noteOn; // polyfill for old browsers
-    audio.currentMusic.start(0);
-    const currentMusicTimeout = setTimeout(() => {
+    if (!buffer) {
+      return console.log('Skipping playing audio ' + file + ', buffer does not exist');
+    }
+
+    audio.gainNode = audio.context.createGain();
+    audio.gainNode.connect(audio.context.destination);
+    audio.gainNode.gain.linearRampToValueAtTime(0, audio.context.currentTime); // Set the node gain to 0 now so that we can fade in soon
```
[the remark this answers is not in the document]
toddmedema: Good call, done

app/actions/Audio.tsx
toddmedema: Done

app/components/Settings.tsx
```diff
@@ -70,6 +71,10 @@ const Settings = (props: SettingsProps): JSX.Element => {
         </div>
       </Picker>
 
+      <Checkbox label="Sound" value={props.audioEnabled} onChange={props.onAudioChange}>
```
smartin015: We should probably also gate this on `NODE_ENV` (just like the test audio buttons) until we're ready to play music in published versions of the app.

app/components/base/Audio.tsx
```diff
@@ -0,0 +1,220 @@
+import * as React from 'react'
+import {loadAudioLocalFile} from '../../actions/Audio'
+import {AudioState, SettingsType} from '../../reducers/StateTypes'
+import {getWindow} from '../../Globals'
+import {logEvent} from '../../Main'
+
+
+const SFX_FILES = ['sfx_combat_defeat', 'sfx_combat_victory'];
```
smartin015: Should these and `MUSIC_FILES` go in `constants.tsx`? Seems like they're used in various places, probs best to reference the same string.

app/components/base/Audio.tsx
```diff
@@ -0,0 +1,220 @@
...
+    this.musicSourceNode.buffer = buffer;
+    this.musicSourceNode.connect(this.musicGainNode);
+    this.musicSourceNode.start = this.musicSourceNode.start || this.musicSourceNode.noteOn; // polyfill for old browsers
+    this.musicSourceNode.start(0);
+    this.musicTimeout = setTimeout(() => {
+      this.playMusicLoopNext();
+    }, (buffer.duration - MUSIC_FADE_SECONDS) * 1000);
+  }
+
+  // TODO nicer implementation would be to save the current spot in the audio (how?)
```
smartin015: RE: how,

`[AudioBufferSourceNode.start()](https://developer.mozilla.org/en-US/docs/Web/API/AudioBufferSourceNode/start)` has an `offset` property, and `[AudioContext.currentTime](https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/currentTime)` is the current hardware time seen by the audio engine. As long as we keep track of the onset of the most recent music file, we can re-start it at `currentTime - startTime`.

With some fades, of course.

app/components/base/Audio.tsx
toddmedema: The next iteration of this PR will be integrating audio into the rest of the code, so this'll naturally find its place as part of that

app/components/Settings.tsx
toddmedema: Good call

smartin015:
This broke tests (discovered in #494)... we should wait for unit tests to pass before merging PRs to master.

PR #482, opened 2017-09-29
allow combat gotos that lead outside of combat

toddmedema:
Previously, a **goto ID** (such as what exists in the dev test quest) would just be ignored and it would proceed to roll and resolve. Not what the writer intends (and have received emails asking to fix this)

My only concern is that this doesn't explicitly clear the combat state or define it as win / lose.... but, I think that might also be ok? For example, it could then allow a quest creator to present a "retreat" option, and the roleplaying card that appears afterwards could access information about the combat (such as duration or number of enemies remaining) to inform what happens during the retreat.


app/cardtemplates/combat/Actions.tsx
```diff
@@ -230,21 +230,24 @@ export function midCombatChoice(settings: SettingsType, parent: ParserNode, inde
...
-      next = next.handleTriggerEvent();
-
-      // If the trigger exits via the win/lose handlers, load it as normal.
-      // Otherwise, we're still in combat.
-      const parentCondition = next.elem.parent().attr('on');
-      if (parentCondition === 'win' || parentCondition === 'lose') {
-        return dispatch(loadNode(settings, next));
+    if (tag === 'trigger') {
+      if (next.elem.text().toLowerCase().startsWith('goto')) {
+        return dispatch(loadNode(null, node.handleAction(index)));
```
smartin015: If we're returning, we don't need an `else` clause.

app/cardtemplates/combat/Actions.test.tsx
```diff
@@ -327,4 +327,6 @@ describe('Combat actions', () => {
   it('handles global player count change');
 
   it('clears combat state on completion');
+
+  it('handles gotos that point to outside of combat');
```
smartin015: - If we're persisting combat state, do roleplay cards properly clear combat settings on load, or do they have the same dark background as in-combat roleplay cards?
- What happens if we jump into a roleplay card in the `on round` handler of another combat node? In the same combat node?

I'd suggest adding (and completing) tests for both of these.

app/cardtemplates/combat/Actions.tsx
toddmedema: Done

app/cardtemplates/combat/Actions.test.tsx
toddmedema: Done

- They maintain combat state, but revert to the light / paper theme
- Added some more test cases and stubs

app/cardtemplates/combat/Actions.test.tsx
```diff
@@ -322,6 +324,26 @@ describe('Combat actions', () => {
...
+      expect(store.getActions()[1].node.elem.text()).toEqual('Outside Roleplay');
+    });
+
+    it('handles GOTOs that point to other roleplaying inside of the same combat', () => {
+      const store = mockStore({});
+      store.dispatch(midCombatChoice(TEST_SETTINGS, newCombatNode(), 5));
+      expect(store.getActions()[1].node.elem.text()).toEqual('rp2');
+    });
+
+    it('throws an error if GOTO points to inside of a different combat');
```
smartin015: AFAIK this isn't current behavior - it would just do it and break the quest horribly, probably by treating the destination combat as the start combat. Did you observe what happened when this is tried?

app/cardtemplates/combat/Actions.test.tsx
```diff
@@ -322,6 +324,26 @@ describe('Combat actions', () => {
...
+
+    it('renders light theme class if follows a GOTO to an RP outside of combat');
+
+    it('renders dark theme class for RPs inside of same combat');
```
smartin015: I think these are actually worth writing out before the PR lands... this behavior is weird and very likely to bite us in strange ways later.

app/cardtemplates/combat/Actions.test.tsx
toddmedema: Come to think of it, this would be far better in crawler code. 100% better to catch automatically while writing than breaking a user's quest after they've invested 2 hours into it.

Rm'd this test, added issue for QDL: https://github.com/ExpeditionRPG/expedition-qdl/issues/20

app/cardtemplates/combat/Actions.test.tsx
toddmedema: Went down this rabbit hole for a while, but couldn't figure out a way to get the output `<div>` without having to mock a huge amount of the state (specifically, the below `/cardtemplates/combat/combat` state) to be able to call `Combat(props)` and get a rendered result. Open to ideas; not sure if re-architecting the combat card for the sake of two tests is worth it right now.

The best thing I can think of would be to specifically export `renderMidCombatRoleplay` with its own, minimal set of props... but then we have to maintain that set of props differently than every other combat card, so not sure it's worth the headache.

Another route would be to inspect the state to see if it looks like a combat card vs looks like a roleplay card, but at that point it seems so fragile that it'd hardly be worth writing the test. (and since that still wouldn't guarantee the thing it's actually supposed to be testing)

```
export interface CombatStateProps extends CombatState {
  card: CardState;
  settings: SettingsType;
  maxTier?: number;
  node: ParserNode;
  roundTimeMillis: number;
  victoryParameters?: EventParameters;
}

export interface CombatDispatchProps {
  onNext: (phase: CombatPhase) => void;
  onDefeat: (node: ParserNode, settings: SettingsType, maxTier: number) => void;
  onRetry: () => void;
  onVictory: (node: ParserNode, settings: SettingsType, maxTier: number) => void;
  onTimerStop: (node: ParserNode, settings: SettingsType, elapsedMillis: number, surge: boolean) => void;
  onReturn: () => void;
  onTierSumDelta: (node: ParserNode, current: number, delta: number) => void;
  onAdventurerDelta: (node: ParserNode, settings: SettingsType, current: number, delta: number) => void;
  onEvent: (node: ParserNode, event: string) => void;
  onCustomEnd: () => void;
  onChoice: (settings: SettingsType, parent: ParserNode, index: number) => void;
  onSurgeNext: (node: ParserNode) => void;
}
```

smartin015: What if we spied on `[renderCardTemplate](https://github.com/ExpeditionRPG/expedition-app/blob/master/app/cardtemplates/Template.tsx#L31)` and checked if the returned card was combat or roleplay?

Issue #328, opened 2017-05-06
Catch + report all errors


toddmedema:
Use window.onerror http://stackoverflow.com/questions/951791/javascript-global-error-handling

Add test cases!

Show toast to user and send to Google Analytics, and add to Todd weekly review to check app error events (which may, of course, prompt us to add more errors, more specificity, etc as needed)

Note that this is especially important for quest writers, who frequently encounter and accidentally create crashing bugs

PR #490, opened 2017-10-05
indicate to users when an error happens + automatically report

toddmedema:
Depends on https://github.com/ExpeditionRPG/expedition-qdl/pull/19

(so once that gets to deployed to master, we'll need to update package.json here)

Closes https://github.com/ExpeditionRPG/expedition-app/issues/328


app/Main.tsx
```diff
@@ -178,12 +179,27 @@ function setupGoogleAnalytics() {
...
-  console.log('google analytics set up');
 }
 
 export function init() {
-  // Setup as web platform as default; we might find out later we're an app
   const window = getWindow();
+
+  // Catch and display / log all errors
+  // TODO PROBLEM
+  // Once the parsing error is encountered, it breaks all future attempts to deal with store
```
smartin015: Add a reference to #491

app/Main.tsx
```diff
@@ -178,12 +179,27 @@ function setupGoogleAnalytics() {
...
+  // Render, etc
+  window.onerror = function(message: string, source: string, line: number) {
+    const quest = getStore().getState().quest;
+    if (quest && quest.details && quest.details.id) {
+      message = `Quest: ${quest.details.id} - ${quest.details.title}. Error: ${message}`;
+    }
+    const label = (source) ? `${source} line ${line}` : null;
+    logEvent('APP_ERROR', {action: message, label});
+    getStore().dispatch(openSnackbar('Error encounter! Reported automatically.'));
```
smartin015: "Error encountered"? and "Reported" might make people think we're reporting them.

toddmedema: We are - it's getting automatically logged to Google Analytics, including the quest name + id. That being said, would still be great to engage with the person, at the very least to let them know when it's fixed - updated to `Error! Please report`

app/Main.tsx
toddmedema: Woops! That's totally not a bug any more. Rm'd comment and closed issue, sorry for the confusion

app/Main.tsx
```diff
@@ -178,12 +179,24 @@ function setupGoogleAnalytics() {
...
+
+  // Catch and display + log all errors
+  window.onerror = function(message: string, source: string, line: number) {
+    const quest = getStore().getState().quest;
+    if (quest && quest.details && quest.details.id) {
+      message = `Quest: ${quest.details.id} - ${quest.details.title}. Error: ${message}`;
+    }
+    const label = (source) ? `${source} line ${line}` : null;
+    logEvent('APP_ERROR', {action: message, label});
+    getStore().dispatch(openSnackbar('Error! Please report.'));
```
smartin015: Maybe we can direct them to the "send feedback" button in the app? Otherwise we might just confuse people.

toddmedema: Changed to "Please send feedback." and added console.error

PR #495, opened 2017-10-07
Remote play: intercept and display move events

smartin015:
multitouchtest


smartin015:
New approach:

remotebuttons

Also included are some safeguards for combat state (comes in handy when trying to start from cursor in the QC inside an interstitial roleplay card)

Issue #446, opened 2017-09-10
Remote play demo


smartin015:
Colors from https://github.com/ExpeditionRPG/expedition-app/issues/472
![30487711-40353e44-9a02-11e7-8396-9d1cd2e66cc2](https://user-images.githubusercontent.com/775657/30522274-62bc83f6-9b9b-11e7-9145-c8206cbe86bd.png)

toddmedema:
Colors from https://github.com/ExpeditionRPG/expedition-app/issues/472
![30487711-40353e44-9a02-11e7-8396-9d1cd2e66cc2](https://user-images.githubusercontent.com/775657/30522274-62bc83f6-9b9b-11e7-9145-c8206cbe86bd.png)

smartin015:
Suggestion from #475:

<em>Idea from last night: each action includes a hash of the current state (not the entire state, since that would include local settings - perhaps just the hash of all history.action_names? or the hash of the currently displayed HTML? or even just history.length?). The receiving client hashes its own state before running the new action to confirm that it's in sync

If the receiving client recognizes it's out of sync, it requests a force sync, which would either be downloading all of the current state, downloading all past actions, or some other mechanism by which a client that gets disconnected / is in an untrusted / corrupted state can resume playing.</em>

smartin015:
Action items:
- [X] Rewritten dispatch flow for async hooks
- [X] Synced basic navigation
- [ ] Improved test toolkit (automatically sets up remoteify reducer stuff)
- [ ] Rewritten non-remote actions (dialogue, announcement, settings, snackbar, user, feedback)
- [ ] Console warnings when actions not using `remoteify`
- [ ] Synced quest loading
- [ ] Synced roleplay actions
- [ ] Synced combat actions
- [ ] Finalized lobby UI
- [ ] Finalized remote UI overlay footer
- [ ] Finalized basic network code & authentication
- [ ] Alpha version published, feedback gathered
- [ ] Iteration on alpha feedback
- [ ] Network security review & load test
- [ ] Larger closed beta version published, feedback gathered
- [ ] Iteration on beta feedback
- [ ] Open beta

smartin015:
Design decisions:

- We'll allow async actions, but rewind the user when another transaction supersedes theirs.
- This requires a sharded counter per session that allows for ordering events.
- Napkin math indicates that at a 3x sharded counter and 100 remote play quests per day, we'd be hard pressed to go over even $4 a month in Firestore quota.
[...]

PR #497, opened 2017-10-08
Remote Play: Support quest loading, roleplay, (buggy) combat

smartin015:
Required splitting logic for saving state to history away from card navigation :(

On the bright side, "when stuff saves" is much more explicit now and can be done arbitrarily.

This also solidifies behavior of the remote play wrapper around actions - the invocation of the `remotify`'d action is pushed to remote clients, but *all* actions dispatched as a result of the invocation remain local. Any actions not wrapped in `remotify` are treated as local.


smartin015:
Ready for review!

smartin015:
Also, dis related to #446

PR #498, opened 2017-10-10
RemotePlay random seeding

smartin015:
This PR makes combat damage, rolls, & loot the same across clients by passing a random seed when invoking remotified actions.

Also some small fixes for GM mode, which didn't quite work remotely before (the usual "sticky history" issue).


smartin015:
Related to #446 btw

PR #499, opened 2017-10-12
expedition-quest-creator #419 - add missing svg icons

ArcanaIreanol:
(copy paste) WARNING: some icons are small ones (kind of bold)

Closes https://github.com/ExpeditionRPG/expedition-quest-creator/issues/419


toddmedema:
@ArcanaIreanol thanks for this! Where are you copying and pasting the icons from? If you're looking for the not-small version of icons, here's the master repository of icons: https://github.com/ExpeditionRPG/expedition-art/tree/master/icons

If you can update the icons you're warning about to use the not small versions, we can merge this in :)

(We have an outstanding issue to switch the app to pull directly from the art repo, but that will take some time and this is a great temporary solution)

ArcanaIreanol:
I copied from the very same folder (basically renaming). That's why some icons are kind of bold, that's because it did not existed in that folder in non small

I'll recommit tomorrow then. Thanks for the hint

ArcanaIreanol:
FYI: medical/caduceus icon doesn't exists on art repo

PR #505, opened 2017-10-18
add support for full- and half-width art; fix background so that it is fixed to text / scrolls with text

toddmedema:
Deployed to beta so that you can check out the new background scroll behavior :)

screen shot 2017-10-18 at 12 44 15
screen shot 2017-10-18 at 12 44 04


app/cardtemplates/roleplay/Roleplay.tsx
```diff
@@ -23,11 +23,21 @@ export interface RoleplayDispatchProps {
...
+// Replaces :icon_name: with <img class="inline_icon" src="images/icon_name_small.svg">
+// And [art_name] with <img class="art" src="images/art_name.svg">
+// if [art_name] ends will _full, adds class="full"; otherwise displays at 50% size
 function generateIconElements(content: string): string {
-  return content.replace(REGEX.ICON, (match:string, group:string): string => {
-    return `<img class="inline_icon" src="images/${group}_small.svg">`;
-  });
+  return content
+      .replace(REGEX.ICON, (match:string, group:string): string => {
+        return `<img class="inline_icon" src="images/${group}_small.svg">`;
```
smartin015: Any chance we could *not* do this via HTML injection?

toddmedema: I spent several hours attempting to do that for another issue - inserting the icons as SVGs so that we could color / style them - and was not able to make any progress due to how we construct the card contents in several passes. Given how much we already have on our plate, that seems like pretty low reward / high effort work.

toddmedema:
@smartin015 updated with the JSX improvement as discussed :)

PR #506, opened 2017-10-18
audio 2.0: treat music as layers

toddmedema:
Enough of the audio component has changed that it's probably just easiest to read the file / not deltas


app/components/base/Audio.tsx
```diff
@@ -3,11 +3,74 @@ import {loadAudioLocalFile} from '../../actions/Audio'
...
+    theme: 'combat_intense',
+    tracks: [0, 2, 3],
+  },
+  8: {
+    theme: 'combat_intense',
+    tracks: [0, 1, 2, 3],
+  },
+} as {[key: number]: {theme: string, tracks: number[]}};
+
+type gainNode = any;
```
smartin015: Can we CapitalCase these types? otherwise they look like variable names.

app/components/base/Audio.tsx
```diff
@@ -136,20 +202,37 @@ export default class Audio extends React.Component<AudioProps, {}> {
...
+    if (fadeIn) {
+      gain.gain.linearRampToValueAtTime(1.0, this.ctx.currentTime + MUSIC_FADE_SECONDS);
     }
     const source = this.ctx.createBufferSource();
     source.buffer = this.buffers[file];
-    source.connect(this.ctx.destination);
-    source.start = source.start || (source as any).noteOn; // polyfill for old browsers
+    source.connect(gain);
+    source.start = source.start || (source as sourceNode).noteOn; // polyfill for old browsers
     source.start(0);
```
smartin015: Should we align this start point more explicitly (that is, should we be calculating a hard time starting point some time after this.ctx.currentTime)? Otherwise, we may encounter slew in audio timing (especially between tracks in the same measure/section) based on how real-timey the JS client is.

app/components/base/Audio.tsx
toddmedema: Done

app/components/base/Audio.tsx
toddmedema: Just tested on several slower devices around the house. Although pages lagged (per usual) and the audio occasionally crackled, I never experienced out of sync issues.

PR #508, opened 2017-10-20
change 'easy' to 'story' to be less condescending

toddmedema:


toddmedema:
(note: tests will be fixed when it gets merged into master / already solved on master)

PR #510, opened 2017-10-28
fix google analytics

toddmedema:


app/Main.tsx
```diff
@@ -16,7 +16,7 @@ import {openSnackbar} from './actions/Snackbar'
 import {silentLogin} from './actions/User'
 import {handleRemotePlayEvent} from './actions/RemotePlay'
 import {getStore} from './Store'
-import {getWindow, getGapi, getGA, getDevicePlatform, getDocument, setGA, setupPolyfills} from './Globals'
+import {getAppVersion, getWindow, getGapi, getGA, getDevicePlatform, getDocument, setGA, setupPolyfills} from './Globals'
```
smartin015: Why not just import a constant value from `./Constants`?

app/Main.tsx
```diff
@@ -138,7 +150,7 @@ function setupEventLogging() {
     });
   } else {
     window.FirebasePlugin = {
-      logEvent: (name: string, args: any) => { console.log(name, args); },
+      logEvent: (name: string, args: any) => { console.info('Firebase log: ', name, args); },
```
smartin015: If this is done *instead* of firebase logging, we should probably indicate so in the console print.

app/Main.tsx
```diff
@@ -152,35 +164,19 @@ function setupHotReload() {
   }
 }
 
+// disabled during local dev
```
smartin015: Can we avoid initialization entirely when we're running in dev mode?

app/Main.tsx
toddmedema: `getAppVersion()` is used in a number of places; it pulls from package.json, and prevents us from having to define the app version in two places. I suppose we could also move getAppVersion to an APP_VERSION constant that equals `PACKAGE.version`, though I suspect that might break tests / might not be worth the hassle

app/Main.tsx
toddmedema: Good call

app/Main.tsx
toddmedema: Fair; had to undisable it for testing, but should return to being totally disabled. Fixed.

PR #511, opened 2017-10-30
Basic remote play (using Firestore)

smartin015:
Part of #446, this introduces a number of critical changes:

- We're now using Cloud Firestore for remote play event dispatch. This is more reliable overall than hosting our own websocket server, but has issues serving high-frequency updates (read: `touchmove` events). 
- Refactor of remote touch events using `RemoteAffector` (handles event logic) and `RemoteRipple` (wraps JSX elements to produce a ripple effect when remote client interacts with it).


webpack.dist.config.js
```diff
@@ -30,9 +30,9 @@ const options = {
         // Default to beta for safety
         'NODE_ENV': JSON.stringify(process.env.NODE_ENV || 'dev'),
         'API_HOST': JSON.stringify(process.env.API_HOST || 'http://betaapi.expeditiongame.com'),
+        'SERVICE_HOST': JSON.stringify(process.env.SERVICE_HOST || 'http://betaservice.expeditiongame.com'),
       },
     }),
-    new webpack.optimize.ModuleConcatenationPlugin(),
```
toddmedema: Why're we removing this?

toddmedema: (specifically, since removing this, have you confirmed that building the app to beta still works?)

webpack.dist.config.js
```diff
@@ -30,9 +30,9 @@ const options = {
         // Default to beta for safety
         'NODE_ENV': JSON.stringify(process.env.NODE_ENV || 'dev'),
         'API_HOST': JSON.stringify(process.env.API_HOST || 'http://betaapi.expeditiongame.com'),
+        'SERVICE_HOST': JSON.stringify(process.env.SERVICE_HOST || 'http://betaservice.expeditiongame.com'),
```
toddmedema: Please add the SERVICE_HOST definition to the local webpack build file as well :)

smartin015: SERVICE_HOST is actually not used anymore ^^ removed.

webpack.dist.config.js
smartin015: Not sure why this was removed - I added it back in.

Issue #502, opened 2017-10-15
Give an informative error when cookies/site data turned off


toddmedema:
https://www.chromium.org/for-testers/bug-reporting-guidelines/uncaught-securityerror-failed-to-read-the-localstorage-property-from-window-access-is-denied-for-this-document

Bit one of our quest crafters; we could easily catch this error and throw a modal describing what to do to fix it.

Duplicate of https://github.com/ExpeditionRPG/expedition-quest-creator/issues/422, but we should implement for the app as well

PR #520, opened 2017-11-08
alert users when cookies disabled

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/502


app/Globals.tsx
```diff
@@ -131,23 +139,32 @@ export function getNavigator(): any {
   return refs.navigator;
 }
 
+// Can't set it by default, since some browsers on high privacy throw an error when accessing window.localStorage
+export function enableLocalStorage(): void {
+  refs.localStorage = getWindow().localStorage;
+}
+
+export function getLocalStorage(): any {
+  return refs.localStorage;
```
smartin015: I don't think we need to export this quite yet... suggest removing this and just using `refs.localStorage` in the getters.

Issue #513, opened 2017-11-03
Cordova apps error on load


toddmedema:
It's specifically breaking as part of the request for `https://apis.google.com/_/scs/apps-static/_/js/k=oz.gapi.en.lEFZxleuj7s.O/m=auth2,client/rt=j/sv=1/d=1/ed=1/am=EQ/rs=AGLTcCPYEdIaVQLmJ7bpeAmhkuAVoTbh5A/cb=gapi.loaded_0`

But it doesn't break on the web app, which makes me think it might be a Cordova plugin. Could also be related to the recent switch from raw Google Analytics to the ReactGA plugin (less likely, given that that request is for `auth2,client`)

Throws errors on both Android and iOS

toddmedema:
Alright, should be fixed by https://github.com/ExpeditionRPG/expedition-app/commit/9ad82eb8c0b67795bf5f77b05b0af0d52ce4e3ba in the next deploy

toddmedema:
Nevermind - did a final test in preparation for deploy, and looks like it's still throwing an error :(

Additional observation: the 400 error occurs whether the user is logged in or not

toddmedema:
Marking this as high priority because having the first thing thousands of users see when opening the app be "Error!" is not a great experience....

toddmedema:
Observation: It appears to occur when gapi.auth2 is called on a cordova app. Which, is strictly unnecessary since Cordova apps have their own authentication plugin.

I'm going to move gapi initialization into auth since it's only used there, and only have it initialize on a call to webLogin. That should sort this out!

Issue #503, opened 2017-10-18
Improved large screen UI


toddmedema:
Inspiration: http://store.steampowered.com/app/711240/Grand_Academy_for_Future_Villains/

Basically just do what we do now, but without the arbitrary border on the app; instead, limit the text / button content width inside of the app

PR #521, opened 2017-11-09
better large screen / tablet UI

toddmedema:
Most notable improvement is that now the touch area is the full size of the screen :)

screen shot 2017-11-08 at 19 53 48
screen shot 2017-11-08 at 19 53 25
screen shot 2017-11-08 at 19 53 14

Closes https://github.com/ExpeditionRPG/expedition-app/issues/503
Closes https://github.com/ExpeditionRPG/expedition-app/issues/513


app/Main.tsx
```diff
@@ -180,6 +180,9 @@ export function init() {
   const window = getWindow();
   const document = getDocument();
 
+  // Before error handling; being offline / not fetching announcements shouldn't display error message
+  getStore().dispatch(fetchAnnouncements());
```
smartin015: I think it'd be safer to add a try/catch within fetchAnnouncements() that ensures all errors are suppressed. That way, we don't have to worry about the ordering of this code.

app/Style.scss
```diff
@@ -1134,29 +1141,11 @@ $fontsizemap: (
 
   .app_container {
     position: relative;
-    border-left: $border_desktop;
-    border-right: $border_desktop;
     display: block;
-    overflow: hidden;
-    width: $abswidthmax;
-    height: $absheightmax;
+    width: 100%;
```
smartin015: Not that I expect things to explode, but... have you tried this out on regular web and portrait-mode android devices?

app/Main.tsx
toddmedema: So, retried this - and turns out it's not actually an issue / it's a different part of the code throwing that error. See: https://github.com/ExpeditionRPG/expedition-app/issues/513

app/Style.scss
toddmedema: Yep!

PR #531, opened 2017-11-26
Add _.currentCombatTier(), _.aliveAdventurers()

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/517

All things requested by authors. Only question I can think of is, are these the best function names? We've never really discussed a naming pattern for scope functions...

### DIFF
--- a/app/cardtemplates/Template.tsx
+++ b/app/cardtemplates/Template.tsx
@@ -75,6 +75,10 @@ export function defaultContext(): TemplateContext {
         const settings = getStore().getState().settings;
         return settings && settings.contentSets;
       },
+      hasTheHorror: function(): boolean {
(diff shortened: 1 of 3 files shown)


app/cardtemplates/Template.tsx
```diff
@@ -75,6 +75,10 @@ export function defaultContext(): TemplateContext {
         const settings = getStore().getState().settings;
         return settings && settings.contentSets;
       },
+      hasTheHorror: function(): boolean {
```
smartin015: I don't think we should let people programmatically do things based on the expansion sets an app has listed... this breaks the abstraction of "only quests published that require The Horror actually use horror cards/idioms".

PR #534, opened 2017-12-03
new audio engine

toddmedema:


app/actions/Card.tsx
```diff
@@ -25,8 +26,10 @@ export const toCard = remoteify(function toCard(a: ToCardArgs, dispatch?: Redux.
   if (!a.noHistory) {
     dispatch({type: 'PUSH_HISTORY'});
   }
+  if (state.audio.intensity > 0 && (a.name !== 'QUEST_CARD' || a.phase === null || a.phase === 'ROLEPLAY')) {
```
smartin015: Why is intensity set in the card? Could we instead do this in the Combat or Audio JSX element?

app/cardtemplates/combat/Actions.tsx
```diff
@@ -59,6 +61,10 @@ export const initCustomCombat = remoteify(function initCustomCombat(a: InitCusto
   return {};
 });
 
+function getAudioIntensity(currentTier: number, maxTier: number, deadAdventurers: number, roundCount: number): number {
+  return Math.round(Math.min(36, 2 * currentTier + 1 * maxTier + 4 * deadAdventurers + 0.5 * roundCount));
```
smartin015: Pls add comment describing magic numbers (or better, constant values)

app/cardtemplates/combat/Actions.tsx
```diff
@@ -59,6 +61,10 @@ export const initCustomCombat = remoteify(function initCustomCombat(a: InitCusto
   return {};
 });
 
+function getAudioIntensity(currentTier: number, maxTier: number, deadAdventurers: number, roundCount: number): number {
```
smartin015: What about `calculateAudioIntensity`? This doesn't get an existing value, but rather makes one.

app/actions/Card.tsx
toddmedema: So, my line of thinking was that there are a lot of ways to get out of combat, and if we fail at catching even one of them, their app will be stuck playing music forever. So an over-aggressive, global catch seemed like a pretty good solution from a user experience perspective. Open to other ways we could achieve that :) Perhaps by passing node information to the audio element, and having it do this check?

app/cardtemplates/combat/Actions.tsx
toddmedema: 👍

app/cardtemplates/combat/Actions.tsx
toddmedema: 👍

app/actions/Card.tsx
toddmedema: (fixed - moved to audio element)

PR #537, opened 2017-12-05
fix custom combat defeat bug + tier / adventurer bug

toddmedema:


toddmedema:
@smartin015 updated!

toddmedema:
@smartin015 I attempted to use the willLeave() / didLeave() system, but it apparently requires you to use the vanilla TransitionGroup instead of the CSSGroup that we're currently using, which... pretty much broke everything.

app/cardtemplates/combat/CombatContainer.tsx
```diff
@@ -53,7 +53,7 @@ const mapStateToProps = (state: AppStateWithHistory, ownProps: CombatStateProps)
     ...combat,
     card: ownProps.card,
     settings: state.settings,
-    node: ownProps.node, // Persist state to prevent sudden jumps during card change.
+    node: state.quest.node,
```
smartin015: Can you also do this on roleplay card, just for consistency?

PR #540, opened 2017-12-07
Handle resolution of rejected transactions

smartin015:
UI sort of works - there's a bit of glitching towards the end of the "syncing" page as it fades out. But more importantly, it actually handles rejected transactions and reverts to known good state.

#446


app/Style.scss
```diff
@@ -64,7 +64,7 @@
 }
 .fade-leave.fade-leave-active {
   opacity: 0.01;
-  transition: opacity 300ms ease-in;
+  transition: opacity 500ms ease-out;
```
toddmedema: Does this change the existing card transition timing, or is this only used for remote play?

app/components/base/remote/RemoteSync.tsx
```diff
@@ -10,6 +10,9 @@ import CircularProgress from 'material-ui/CircularProgress';
 
 const ReactCSSTransitionGroup: any = require('react-addons-css-transition-group');
 
+export const FADE_ENTER_ANIMATION_MS = 500;
+export const FADE_LEAVE_ANIMATION_MS = 500;
```
toddmedema: I'm not sure why GitHub doesn't think this is going to be a merge conflict, but this seems like it directly conflicts with the code we added here, to time card rendering based on transitions: https://github.com/ExpeditionRPG/expedition-app/pull/537/files#diff-026d79760ed2f976b42ba691c664042aR145

Probably easiest to delete this and just use the Global constant CARD_TRANSITION_ANIMATION_MS defined in the other PR, though it looks like we will need to update its value to 500

toddmedema: Oh wait. This is just for the remote play "Syncing..." spinner. OK, so this probably won't be a conflict. But, it might make sense to compact / unify their implementation somewhat? I'll leave that to your discretion.

app/Style.scss
smartin015: Just for remote play

app/components/base/remote/RemoteSync.tsx
smartin015: Let's keep them separate for now since fading is a bit slower of an animation.

PR #542, opened 2017-12-08
re-add pay what you want

toddmedema:
This will help give us a sense of if our fans are willing to pay additional money once they have the game (for example, to purchase voice-over quests).

Also, if this does work out, we'll gain a recurring revenue stream + provide a lot more motivation to authors. So it's a win win :)

Note that this is based on previous work from the summer; we'd originally abandoned it because it seemed like a lot of work was remaining... but that was just feature creep IMHO, and we're better off launch this now / soon and rolling out improvements as needed.

screen shot 2017-12-08 at 11 53 02

screen shot 2017-12-08 at 12 07 55
screen shot 2017-12-08 at 12 08 09
screen shot 2017-12-08 at 12 08 17


app/components/Checkout.tsx
```diff
@@ -0,0 +1,111 @@
...
+  constructor(props: CheckoutProps) {
+    super(props);
+    this.state = { card: null, paymentError: null, paymentValid: false };
+  }
+
+  componentDidMount() {
+    if (this.state.card) { return; }
+    let stripe = this.props.checkout.stripe;
+    if (!stripe) {
+      stripe = Stripe(authSettings.stripe);
```
smartin015: Are we sure the Stripe API is OK with multiple instances of stripe object? Why do we need this in addition to the one in redux store?

app/components/Checkout.tsx
```diff
@@ -0,0 +1,111 @@
...
+    const elements = stripe.elements();
+    const card = elements.create('card', {
+      style: {
+        base: {
+          fontSize: '18px',
+          fontFamily: 'MinionPro, serif',
+        },
+      },
+    });
+    card.mount('#stripeCard');
```
smartin015: Do we need to unmount this when the react component dies?

app/components/CheckoutContainer.tsx
```diff
@@ -0,0 +1,74 @@
...
+    },
+    onPhaseChange: (phase: CheckoutPhase): void => {
+      dispatch(checkoutSetState({phase}));
+    },
+    onStripeLoad: (stripe: any): void => {
+      dispatch(checkoutSetState({stripe}));
+    },
+    onSubmit: (stripeToken: string, checkout: CheckoutState, user: UserState): void => {
+      dispatch(checkoutSetState({phase: 'PROCESSING'}));
+      fetch(authSettings.urlBase + '/stripe/checkout', {
```
smartin015: This should be in `actions/Checkout.tsx`.

app/components/CheckoutContainer.tsx
```diff
@@ -0,0 +1,74 @@
...
+        method: 'POST',
+        body: JSON.stringify({
+          token: stripeToken,
+          amount: checkout.amount,
+          productcategory: checkout.productcategory,
+          productid: checkout.productid,
+          userid: user.id,
+          useremail: user.email,
+        }),
+        // mode: 'cors',
```
smartin015: RM dead code

app/components/QuestEnd.tsx
```diff
@@ -32,6 +34,8 @@ export default class QuestEnd extends React.Component<QuestEndProps, {}> {
   render() {
     const loggedIn = (this.props.user && this.props.user.loggedIn);
     const rated = (this.props.userFeedback.rating > 0);
+    // TODO develop system for letting the app know when payments are offline
```
smartin015: If the user tries to tip and the system is offline, what happens?

app/components/QuestEnd.tsx
```diff
@@ -54,6 +58,18 @@ export default class QuestEnd extends React.Component<QuestEndProps, {}> {
             />
           </div>
         }
+        Tip the author:
+        <div className={'tipAmounts ' + (checkoutEnabled ? '' : 'checkoutDisabled')}>
+          <Button onTouchTap={() => this.props.onTip(checkoutEnabled, 1, this.props.quest, this.props.user)}>
+            $1
```
smartin015: Are we going to show this in non-US countries? How do we handle the currency conversion?

app/reducers/StateTypes.tsx
```diff
@@ -20,8 +20,16 @@ export interface AudioState {
   timestamp: number;
 }
 
-export type DialogIDType = null | 'EXIT_QUEST' | 'EXPANSION_SELECT';
+export type CheckoutPhase = 'ENTRY' | 'PROCESSING' | 'DONE';
+export interface CheckoutState {
+  amount: number;
+  phase: CheckoutPhase;
```
smartin015: As discussed, let's move this phase section to CardState.phase and make "processing" its own attribute.

app/components/QuestEnd.tsx
toddmedema: Stripe automatically handles currency conversions on a transactional level: https://stripe.com/docs/currencies/conversions

Not worried about displaying internationalized currencies as part of v1, since we don't internationalize any other part of the app.

app/components/Checkout.tsx
toddmedema: So, reviewing the logic flow here, it only instantiates the Stripe object if it doesn't exist in the checkout state; and then, once it's loaded, it saves it to the Redux store. So, there should never be multiple instances, this just prevents the app from loading it before it's needed given the small number of users / app loads that will actually check out

app/components/Checkout.tsx
toddmedema: Seemed to work fine without it, but added just to be sure :)

app/components/QuestEnd.tsx
toddmedema: Just tried: Stripe shows a nice network connectivity error inside of the form

app/reducers/StateTypes.tsx
toddmedema: @smartin015 all done!

PR #545, opened 2017-12-14
Fix errors

toddmedema:
- properly catch fetch() errors
- fix iOS not opening email client for reporting errors
- disabe checkout on iOS app (stripe crashes it....)
- require a user review on all terrible ratings


app/components/QuestEndContainer.tsx
```diff
@@ -62,8 +66,9 @@ const mapDispatchToProps = (dispatch: Redux.Dispatch<any>, ownProps: any): Quest
       dispatch(toPrevious({name: 'FEATURED_QUESTS'}));
     },
     onTip: (checkoutEnabled: boolean, amount: number, quest: QuestState, user: UserState) => {
+      logEvent('tip_start', { value: amount, action: quest.details.title, label: quest.details.id });
       if (!checkoutEnabled) {
-        dispatch(openSnackbar('Error encountered loading checkout'));
+        dispatch(openSnackbar('Tipping is currently disabled :('));
```
smartin015: Users might think this is their fault somehow... we should try to give them an idea as to why it's disabled.

toddmedema: 👍

PR #548, opened 2017-12-16
Remote Play: improved combat behavior, numPlayers in client status

smartin015:
This fixes the inconsistencies in combat timer, loot, level up, and enemy damage due to those methods referencing local player count instead of the total number of players. To fix this, the client player count was added to their remote play `StatusEvent` and broadcast to other clients.

Also fixed the "waiting on X peers" by ignoring disconnected clients in `state.remotePlay.clientStatus`.

#446


app/cardtemplates/combat/Actions.tsx
```diff
@@ -17,45 +17,65 @@ import {getRemotePlayClient} from '../../RemotePlay'
...
-    dispatch(toCard({name: 'QUEST_CARD', phase: 'DRAW_ENEMIES', noHistory: true}));
-    dispatch(audioSetIntensity(calculateAudioIntensity(tierSum, tierSum, 0, 0)));
+  }
+  a.node = a.node.clone();
+  const totalPlayerCount = numLocalAndRemotePlayers(a.settings, getState().remotePlay);
+  a.node.ctx.templates.combat = {
+    custom: a.custom || false,
+    enemies,
+    roundCount: 0,
+    numAliveAdventurers: a.settings.numPlayers, // Local player count
```
toddmedema: We should ideally update this so that your app knows and displays the total # of adventurers in the party, and the # of local adventurers is only used for the purpose of multitouch (and perhaps displaying in the client list, oh John's client has 4 players)

smartin015: Done.

PR #550, opened 2017-12-18
lengthen end of music fade; submit feedback on tip; bump Android min version

toddmedema:


app/components/QuestEnd.tsx
```diff
@@ -59,15 +59,15 @@ export default class QuestEnd extends React.Component<QuestEndProps, {}> {
             />
           </div>
         }
-        Tip the author:
+        Tip the author: (also submits review)
```
smartin015: `Tip the author and submit your review:`?

toddmedema: 👍

PR #553, opened 2017-12-27
in-app feedback dialog

toddmedema:
Better user experience - especially on iOS, where there's no way to go back, so once you open the website to contact us, you have to kill and restart the app...

Note that this also requires an API PR such that the new /feedback/feedback emails us the feedback.

screen shot 2017-12-26 at 21 23 57


toddmedema:
@smartin015 since the original architecture looked good, added a new commit that migrates the rest of the feedback-esque routes to the new dialog system (which, added bonus, got rid of 4 files!)

Let me know if that additional commit looks good to you :)

PR #562, opened 2018-01-03
add dark theme to rp footer

toddmedema:
Also a few other small bugfixes from the RP beta that I forgot to commit


app/Main.tsx
```diff
@@ -46,26 +46,26 @@ function setupRemotePlay() {
...
-      value: args.value || undefined,
+      action: argsInput.action || '',
+      label: argsInput.label || '',
+      value: argsInput.value || undefined,
     });
   }
+
+  const fbp = getWindow().FirebasePlugin;
+  if (fbp) {
+    // Firebase requires that keys are < 40 character, values are < 100 characters
```
smartin015: If you added some constants (e.g. `FIREBASE_MAX_VALUE_LENGTH`), this comment would be unnecessary.

toddmedema: 👍

PR #563, opened 2018-01-03
add quest start/end logging

toddmedema:


app/actions/Web.tsx
```diff
@@ -49,9 +50,11 @@ export const fetchQuestXML = remoteify(function fetchQuestXML(details: QuestDeta
 // for loading quests in the app - Quest Creator injects directly into initQuest
 function loadQuestXML(a: {details: QuestDetails, questNode: Cheerio, ctx: TemplateContext}) {
   return (dispatch: Redux.Dispatch<any>) => {
-    // Quest start is here instead of initQuest because initQuest is also used by the editor
-    // and would over-report.
+    // Quest start logging is here instead of initQuest because initQuest is also used by the editor
+    // and would over-report. And not in the quest start component, because that gets skipped
```
smartin015: Sounds like the quest start component needs a rename... care to put in a TODO?

toddmedema: Applied the change here since it was small, changed from QuestStart to QuestSetup

PR #576, opened 2018-01-19
Add remote play stats dialog and analytics instrumentation

smartin015:
For #446

This'll give us a bit more detail on how far into remote play people get before they run into problems, plus some idea on what kind of problems they're facing. 

I'd like to follow on with additional instrumentation that collects and ships off all the recent errors (e.g. the last 50 or so) that the app has thrown, but figured that'd be better to do in a separate PR.


smartin015:
Beta and prod are updated with the new column, and the API PR is now checked in.

PR #590, opened 2018-02-17
Remoteplay - clean up & stub out unit tests

smartin015:
Preparation for more thorough testing of remote play client/dispatch logic. #446


package.json
```diff
@@ -76,6 +76,7 @@
         "webpack-dev-server": "^2.9.4"
     },
     "dependencies": {
+        "@types/body-parser": "^1.16.8",
```
toddmedema: @types should be a dev dependency

PR #600, opened 2018-02-19
Fix quest feedback & include console logs

### LINKED ISSUE #580: submitUserFeedback user specification may be buggy
[link](https://github.com/ExpeditionRPG/expedition-app/blob/1bde135ee44768709dab6dcddf440a550de03505/app/actions/Web.tsx#L119)

`a.user.id` and other fields are used before the check for `!a.user` is done. I expect we could do something better here.

### LINKED ISSUE #581: Intercept, cache, and report console.log/console.error calls
It'd be super useful to have a handy list of logged console events whenever a user reports an error (e.g. in remote play). AFAICT this would be a matter of wrapping console.log and appending to an array or string every time something is logged to the console. 

We may end up with overly large packet size when we upload though... unsure about what we'd need to do to pass e.g. 8k+ of text.

smartin015:
This PR includes console logs when users report feedback. This should significantly improve our ability to debug based on user feedback.

It also properly sends the quest partition (e.g. "expedition-public") along with the feedback so the API server can correctly look up the published quest and provide additional details via email.

Fixes #581
Fixes #580
Useful for #446


app/Main.tsx
```diff
@@ -1,6 +1,38 @@
...
+};
+{
+  const oldLog = console.log;
+  console.log = (...objs: any[]) => {return logHook(oldLog, objs);};
+
+  const oldWarn = console.warn;
+  console.warn = (...objs: any[]) => {return logHook(oldWarn, objs);};
+
+  const oldError = console.error;
+  console.error = (...objs: any[]) => {return logHook(oldError, objs);};
```
toddmedema: It looks like this removes existing console.log functionality? Wouldn't it be better to have it also still console.log so that it's still possible to debug issues? (or maybe only do that on dev, might slightly improve perf on prod?)

app/Console.tsx
```diff
@@ -0,0 +1,9 @@
+const buffer: string[] = [];
+
+export function logToBuffer(line: string) {
+  buffer.push(line);
```
toddmedema: We should probably implement this as a ring buffer so that 100's or 1,000's of logs doesn't crash older devices from eating up all their memory. The last 50-100 logs should be more than sufficient to diagnose issues.

app/Main.tsx
smartin015: It still logs to the console - that's the call to `f(...objects)` in `logHook`.

app/Console.tsx
smartin015: We probably should also see initial logs in case there's something weird with the user's device or setup... updated to include the first 25 logs and the last 50 logs. Also keeps count of how many logs were omitted in between them.

PR #608, opened 2018-03-06
Refactor remote play event handling; add more checks against message duplicates

smartin015:
#446 

- Message handling is now done in the RemotePlayClient, which is a bit easier to comprehend.
- Handles a few more edge cases, such as when messages are failed due to max retries but the server eventually commits them
- inbound ACTION events now advance the `committedEventCounter`.


package.json
```diff
@@ -108,7 +108,7 @@
     "emitter-steward": "^1.0.0",
     "es6-promise-plugin": "^4.1.0",
     "escape-string-regexp": "^1.0.4",
-    "expedition-qdl": "git+https://github.com/ExpeditionRPG/expedition-qdl.git#d51ab56a2c6d120ba1a5539b9a87c8c1dab4c885",
+    "expedition-qdl": "git+https://github.com/ExpeditionRPG/expedition-qdl.git#85a9274469d0e3adecfc08515ae4a2b5c7350f15",
```
toddmedema: Note that you'll need to merge in the QDL PR and then update this hash again - github / npm install doesn't like hashes that point to deleted branches

smartin015: Done

PR #611, opened 2018-03-16
Remote Play Refactor

smartin015:
- Fixes logic in console log wrapper to not error on null/undefined
- Moves committedEventCounter into redux state (so the counter is tightly tied to the committed value)
- Removes inflight action queue in favor of using RemotePlay retry buffer for tracking inflight actions
- No longer handle INFLIGHT_COMMIT and INFLIGHT_REJECT from server; rather, look for self ID in inbound actions
- Append card name to main container key so MULTI_EVENT actions spanning multiple cards don't stop at the first card
- remove need for handleEvent; put logic in routeEvent instead.
- add redux-devtools-extension to filter out REMOTE_PLAY_CLIENT_STATUS in devtools

#446


app/reducers/InFlight.test.tsx
```diff
@@ -16,42 +16,13 @@ describe('InFlight reducer', () => {
...
-      inflight(stateWithInflight, {type: 'INFLIGHT_REJECT', id: 50} as InflightRejectAction, increment);
-    });
-
-    it('rejects higher-valued uncommitted actions', () => {
-      const newState = inflight(stateWithInflight, {type: 'INFLIGHT_REJECT', id: 1} as InflightRejectAction, increment);
-      expect(newState._inflight).toEqual([]);
-    });
-  });
-
+  /*
```
toddmedema: Wer're commenting out all of these tests? Should we just delete them because they're no longer relevant? Or do they need to be reimplemented?

smartin015: Good call; removed them.

PR #612, opened 2018-03-18
get links to properly open in external browser (for iOS)

toddmedema:


app/Main.tsx
```diff
@@ -106,6 +106,7 @@ export function logEvent(name: string, argsInput: any): void {
 
 function setupDevice() {
   const window = getWindow();
+  window.open = ((window.cordova || {}).InAppBrowser || {}).open;
```
smartin015: This yields undefined if the path doesn't resolve.... suggest something like:

`window.open = .... .open || window.open;`

package.json
```diff
@@ -1,171 +1,173 @@
 {
```
smartin015: Change 4 spaces -> 2 spaces plz

PR #613, opened 2018-03-18
standardize remote play icons to use Expedition icons / art

toddmedema:
Matches with art update here: https://github.com/ExpeditionRPG/expedition-art/commit/1dd6989cd50744c68fa157b56edfc60e2cfbcde2

(note that `npm install` sometimes fails to pickup changes to the art repo; when you pull this down, I'd suggest deleting expedition-art from your node_modules, clearing your npm cache, and then npm install)


toddmedema:
(to clarify - these are the same icons as in the material ui base set; we can always customize them more later, but most immediately this prevents occasional weird styling differences by keeping all icons on the same system)

PR #614, opened 2018-03-18
Remote Play - Session Rejoin

smartin015:
#446

Required some surgery of middleware, appcontainer, and navigation code, whee!

- AppContainer now checks on a timestamp plus key derived from card name, card phase, and quest line. This prevents e.g. exclusion of multiple navigations to QUEST_CARD during fast-forward.
- Updated remote play middleware to chain return statements back to the dispatch() function. 
- Updated `fetchLocal` to return a Promise, which is passed via the chained return to let us act when the fetch is complete.
- MULTI_EVENT response now constructs a chain of promises, which allows "blocking" resolution until async actions like fetchQuestXML complete.
- Finished implementing the "remote play from previous session" code.
- Some small UI fixes (e.g. adding scrollbar to remote play status dialog).

ezgif com-optimize


app/components/RemotePlay.tsx
```diff
@@ -43,9 +43,10 @@ class RemotePlayConnect extends React.Component<RemotePlayProps, {}> {
   }
 
   render() {
+    console.log(this.props.remotePlay.history);
```
toddmedema: rm log?

PR #619, opened 2018-03-19
reduce number of any types; closes #574

### LINKED ISSUE #574: Reduce prevalence of "any" type
*Migrated to expeditionrpg/expedition#230 by [spacetelescope/github-issues-import](https://github.com/spacetelescope/github-issues-import)*

We have [roughly 90 occurrence of the "any" type](https://github.com/ExpeditionRPG/expedition-app/search?utf8=%E2%9C%93&q=%22%3A+any%22&type=) in our code. This basically turns off all compile-time assurances for any code that uses these vars. We should be a bit more principled here and only use "any" when it's absolutely necessary - and even then, with very defensive coding.

toddmedema:


app/reducers/StateTypes.tsx
```diff
@@ -138,7 +143,7 @@ export interface RemotePlaySessionMeta {
 }
 
 export interface RemotePlayState {
-  session: {secret: string, id: SessionID}|null;
+  session: RemotePlaySessionType|null;
```
smartin015: Why not RemotePlaySessionMeta?

toddmedema: Hmm. That didn't really seem like the Meta to me? That's the raw ID information, meta information to me is auxiliary stuff like timestamps and counters

Issue #598, opened 2018-02-19
Linkable quests


smartin015:
Click a url - open that specific quest details page via the app. Need a modal/dialog for selecting the number of players before the quest begins.

v2:
- loading spinner
- Research deep linking for Android + iOS

PR #623, opened 2018-03-22
 on app load, if URL hash, attempt to load and play that quest

toddmedema:
Further progress towards https://github.com/ExpeditionRPG/expedition-app/issues/598

Still have to add the player count dialog, but trying to keep each change smaller and more digestible :)


app/actions/Search.tsx
```diff
@@ -21,48 +21,82 @@ export const search = remoteify(function search(a: SearchSettings, dispatch: Red
...
 
   return a;
 });
+
+// TODO this doesn't handle featured quests since they don't show up in a DB search
+// Perhaps just manually checks against the featured quests array?
+export const searchAndPlay = remoteify(function searchAndPlay(id: string, dispatch: Redux.Dispatch<any>) {
+  const params = {
+    id,
+    partition: 'expedition-public',
```
smartin015: Could we include the partition as part of the URI?

app/actions/Search.tsx
```diff
@@ -21,48 +21,82 @@ export const search = remoteify(function search(a: SearchSettings, dispatch: Red
...
+  dispatch(getSearchResults(params, (quests: QuestDetails[], response: any) => {
+    dispatch({
+      type: 'SEARCH_RESPONSE',
+      quests: quests,
+      nextToken: response.nextToken,
+      receivedAt: response.receivedAt,
+      search: {}, // Don't specify search params because this one's weird and uses ID
+    } as SearchResponseAction);
+    if (quests.length === 0) {
+      alert('Quest not found, returning to home screen.');
```
smartin015: I'd add a TODO for better UX here.

app/actions/Search.tsx
```diff
@@ -21,48 +21,82 @@ export const search = remoteify(function search(a: SearchSettings, dispatch: Red
...
+      dispatch(toCard({name: 'SPLASH_CARD'}));
+    } else {
+      dispatch(viewQuest({quest: quests[0]}));
+      dispatch(toCard({name: 'SEARCH_CARD', phase: 'DETAILS'}));
```
smartin015: Looks like navigate action is also done directly after the other call to `viewQuest` in our code. Think it's worth pushing into `viewQuest`?

app/actions/Search.tsx
toddmedema: Since the private search for authors also uses the search function, we can't hardcode that

app/actions/Search.tsx
toddmedema: Added + matching issue + link

app/actions/Search.tsx
toddmedema: Makes sense to me, done

PR #624, opened 2018-03-22
add raven / sentry monitoring

toddmedema:


app/Main.tsx
```diff
@@ -70,6 +71,11 @@ import thunk from 'redux-thunk'
 const injectTapEventPlugin = require('react-tap-event-plugin');
 const ReactGA = require('react-ga');
 
+Raven.config('https://990df74f1b58424395ec3d3ec6f79b42@sentry.io/420182', {
```
smartin015: Should this be an environment var? Seems config-y.

toddmedema: Good point, moved to Constants alongside the other config-y things

PR #633, opened 2018-03-26
 successfully clear quest ID from url hash when going back via back button and The End

toddmedema:
A little awkward b/c now combinedReducer now affects URL hash (only if the history API is available, so as not to break tests), but it's the only single place that has enough context to make that decision since the previous state when going back to FEATURED_QUESTS card still has a quest defined, since we assign quest before transitioning cards


app/cardtemplates/combat/Actions.tsx
```diff
@@ -55,7 +55,7 @@ export function generateCombatTemplate(settings: SettingsType, custom: boolean,
     roundCount: 0,
     numAliveAdventurers: totalPlayerCount,
     tier: tierSum,
-    roundTimeMillis: settings.timerSeconds * 1000 * (PLAYER_TIME_MULT[totalPlayerCount] || 1),
```
smartin015: Why did this change? Shouldn't the round time limit match the total player count for remote play? Sounds to me like `totalPlayerCount` is what's actually incorrect during 1-2p

app/reducers/CombinedReducers.tsx
```diff
@@ -16,6 +16,8 @@ import {AppStateWithHistory, AppState, AppStateBase} from './StateTypes'
 import {ReturnAction} from '../actions/ActionTypes'
 import {getNavigator} from '../Globals'
 
+const historyApi = (typeof history.pushState !== 'undefined');
```
smartin015: This is only used in one place...suggest moving this to the if statement, rename to `hasURLHistory`.
Also, how is `history` defined in this file? I don't see imports or `declare type` anywhere.

app/reducers/CombinedReducers.tsx
```diff
@@ -42,7 +44,7 @@ function isReturnState(state: AppStateBase, action: ReturnAction): boolean {
 }
 
 export default function combinedReducerWithHistory(state: AppStateWithHistory, action: Redux.Action): AppStateWithHistory {
-  let history: AppStateBase[] = [];
+  let historyArray: AppStateBase[] = [];
```
smartin015: `stateHistory`? `historyArray` isn't that much more descriptive than `history`.

app/reducers/CombinedReducers.tsx
```diff
@@ -92,6 +94,10 @@ export default function combinedReducerWithHistory(state: AppStateWithHistory, a
         pastStateIdx--;
       }
 
+      if (historyApi && (pastStateIdx === 0 || state._history[pastStateIdx-1].quest.details.id === '')) {
+        history.pushState(null, '', '#');
```
smartin015: Not really clear why this conditional exists... does this call to `pushState` remove all URL history? Add a comment?

app/cardtemplates/combat/Actions.tsx
toddmedema: Ah, good point. It's because numLocalAndRemotePlayers was getting capped to a minimum of 2, which made it impossible to detect actual single player. Split that out to numLocalAndRemotePlayers / Adventurers for clarity, confirmed expected behavior.

app/reducers/CombinedReducers.tsx
toddmedema: Done

app/reducers/CombinedReducers.tsx
toddmedema: Added. It does not remove all URL history, that is not possible from JS

app/reducers/CombinedReducers.tsx
toddmedema: Good call, much clearer. It's pulled from the window context / built into the browser (hence the gating on detection)

app/cardtemplates/combat/Actions.tsx
```diff
@@ -125,7 +144,7 @@ function getEnemies(node: ParserNode): Enemy[] {
 }
 
 function generateCombatAttack(node: ParserNode, settings: SettingsType, rp: RemotePlayState, elapsedMillis: number, rng: () => number): CombatAttack {
-  const totalPlayerCount = numLocalAndRemotePlayers(settings, rp);
+  const totalPlayerCount = numLocalAndRemoteAdventurers(settings, rp);
```
smartin015: If we're making a distinction between players and adventurers, this should still be `numLocalAndRemotePlayers` (since the [mult constant](https://github.com/ExpeditionRPG/expedition-app/blob/47acfb7dc3a3956003f6f090d17249f8f73904d9/app/Constants.tsx#L61) has a setting for single-player).

app/reducers/Card.tsx
```diff
@@ -3,15 +3,15 @@ import {CardState} from './StateTypes'
 import {CardTransitioningAction, NavigateAction} from '../actions/ActionTypes'
 import {NAVIGATION_DEBOUNCE_MS} from '../Constants'
 
-const historyApi = (typeof history.pushState !== 'undefined');
+const hasURLHistory = (typeof history.pushState !== 'undefined');
```
smartin015: This is now used in multiple files. If it's simple enough, dedupe it into the if statements, otherwise this should probably live in Constants.tsx. Either way it doesn't need to be a global in the files.

app/cardtemplates/combat/Actions.tsx
toddmedema: Ah, yep!

app/reducers/Card.tsx
toddmedema: Good call; moved to `Globals` since that's more inline with our current style

app/cardtemplates/combat/Actions.tsx
```diff
@@ -47,15 +65,16 @@ export function generateCombatTemplate(settings: SettingsType, custom: boolean,
...
   const totalPlayerCount = numLocalAndRemotePlayers(settings, remotePlay);
 
   return {
     custom: custom,
     enemies,
     roundCount: 0,
-    numAliveAdventurers: totalPlayerCount,
+    numAliveAdventurers: totalAdventurerCount,
     tier: tierSum,
-    roundTimeMillis: settings.timerSeconds * 1000 * (PLAYER_TIME_MULT[totalPlayerCount] || 1),
```
smartin015: Final request: unit test to confirm this works for 1p

Issue #637, opened 2018-03-30
Tipping broken again


toddmedema:
The error I got (at least on web):
Error: Uncaught IntegrationError: Stripe.js must be loaded from js.stripe.com. For more information https://stripe.com/docs/stripe-js/reference#including-stripejs

Make sure to test the fix on Android + iOS as well

PR #638, opened 2018-03-31
fix tiping, make its error detection more robust

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/637


app/components/QuestEnd.tsx
```diff
@@ -43,7 +43,9 @@ export default class QuestEnd extends React.Component<QuestEndProps, {}> {
     const rated = this.props.userFeedback.rating !== undefined && (this.props.userFeedback.rating > 0);
     // TODO ping server to determine if payments are enabled
     // TODO figure out why loading Stripe crashes iOS app
-    const checkoutError = (this.props.platform === 'ios') ? 'Checkout currently disabled on iOS app' : null;
+    let checkoutError = null as string|null;
```
smartin015: `let checkoutError: string|null = null;`

Better to type the definition instead of using the "as" assertion.

app/components/QuestEnd.tsx
```diff
@@ -43,7 +43,9 @@ export default class QuestEnd extends React.Component<QuestEndProps, {}> {
...
+    if (this.props.platform === 'ios') { checkoutError = 'Checkout currently disabled on iOS app'; }
```
smartin015: Expand these please, there's no need to one-line them.

Issue #636, opened 2018-03-30
Fix external links on Android


toddmedema:
Looks like we have to use `navigator.app.loadUrl("http://google.com", {openExternal : true});` for Android - https://stackoverflow.com/questions/17887348/phonegap-open-link-in-browser

Should probably just overwrite window.open with whatever function works best on that platform (incl including _system for iOS)

toddmedema:
To clarify, it looks like `<a>` might already work (like privacy policy), just not the shop link (and anything else that uses that same system)

PR #639, opened 2018-03-31
fix android external links

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/636


app/components/FeaturedQuests.tsx
```diff
@@ -48,11 +49,11 @@ const FeaturedQuests = (props: FeaturedQuestsProps): JSX.Element => {
           <div className="title"><img className="inline_icon" src="images/roll_small.svg"/>Tools</div>
         </div>
       </Button>
-      <a href="#" onClick={() => window.open('https://expeditiongame.com/store?utm_source=app', '_system')} className="linkButton"><Button>
+      <Button onTouchTap={()=>openWindow('https://expeditiongame.com/store?utm_source=app')} remoteID="tools">
```
smartin015: Shouldn't `remoteID="shop"`? Alternatively if you leave this off, we won't display tap ripples from other clients (which kind of makes sense).

package.json
```diff
@@ -9,7 +9,8 @@
     "test": "mocha meta_tests.js && karma start --single-run --browsers PhantomJS",
     "watch:test": "karma start",
     "dev": "webpack-dev-server --progress",
-    "build": "webpack --config ./webpack.dist.config.js --progress"
+    "build": "webpack --config ./webpack.dist.config.js --progress",
+    "build-all": "npm run build && cordova build android && cordova build ios"
```
smartin015: I'm not on a mac, so this won't work for me :(

app/components/FeaturedQuests.tsx
toddmedema: Good call, rm'd

package.json
toddmedema: Yep, that's why there's still the normal `build` :) PITA to type all of that out when I'm testing something against all the devices we support.

Issue #599, opened 2018-02-19
Allow reviewer contact opt-out


smartin015:
We should set up feedback so reviewer contact information is sent to the quest author, but also that the reviewer is made aware of this and can opt out if they want.

This could result in higher-quality reviews (since the reviewer knows they could be contacted) but still allows for anonymous feedback.

PR #641, opened 2018-03-31
Allow anonymous feedback

smartin015:
#599 
image


toddmedema:
Please add some (standard-sized) bottom padding to the Give Feedback checkmark container - right now it's visually closer to "Tip the author" than "Tip your author" is to the tipping values!

smartin015:
Updated, PTAL

PR #648, opened 2018-04-14
add moar tests

toddmedema:


app/cardtemplates/combat/State.test.tsx
```diff
@@ -45,5 +61,38 @@ describe('Combat State', () => {
       it('only includes enemies in the currently enabled content sets');
     });
 
+    describe('aliveAdventurers', () => {
+      it('returns the numer of aliver adventurers when all alive', () => {
```
smartin015: Nit: `alive`, here and next test. I'd suggest making this a stub again (no function param in the call to `it`) since it's not actually a test.

PR #650, opened 2018-04-17
disable search, url hash, remote play in simulator

toddmedema:


app/actions/Card.tsx
```diff
@@ -24,7 +24,7 @@ export const toCard = remoteify(function toCard(a: ToCardArgs, dispatch: Redux.D
       nav.vibrate(VIBRATION_SHORT_MS);
     }
   }
-  if (!a.noHistory) {
+  if (!state.settings.simulator && !a.noHistory) {
```
smartin015: Why not push history? This will break back-arrow functionality in the quest creator AFAICT.

toddmedema: Fixed

Issue #466, opened 2017-09-13
Saving spot in quest


toddmedema:
As talked about here: https://github.com/ExpeditionRPG/expedition-app/pull/465 - now that we have access to local storage, we can save and load quest state strings.

Thoughts / questions:

- Initially limiting to one quest save at a time should be fine for the majority case. Especially given the limitations around storage size, and the fact that iOS can wipe storage at any time, makes me think that storing multiple quest states might require some pretty significant hijinks.
- Could save as two key, values: quest id and the state string? Maybe also name too, so that even if you're offline, the app can still show "resume playing QUEST NAME"
- Question: Is state serialization working? What does that look like?
- Question: how big are serialized state strings? (localstorage is limited to 5MB). Specifically, how does the length of the string grow with the size of the quest, and with how far players have played into the quest. Note that for the sake of storage, I think it would be reasonable to only save the current state / not the entire history.
- Question: How would the load / resume function work? Presumably the app would need to have access to the quest XML. Perhaps we are also able to fit the XML in the localstorage 5MB, as a first implementation of offline quests? (ie if you start playing it, you can keep playing it even w/o a connection)
  - Mistress Malaise XML is only 17kb. Doubled for string encoding, we could still fit a single quest that's 140x longer than MM (aprox 32,000 lines of XML) into 5MB local storage.

toddmedema:
Another possible implementation would be to use this cordova SQLlite plugin, which (as far as we can tell) does not have any storage limits: https://github.com/litehelpers/Cordova-sqlite-storage

Of course, there end up being numerous architectural challenges, including how to handle migrations of saves from old engine versions when we deploy updates (to which Scott suggests running the same play-path through the crawler to re-generate the current save position) and handling updated versions of the quest (to which I suggest saving the version of the quest you're using locally, and only playing off that / not allowing it to be updated mid-save)

Many of these things will end up being solved (or at least refined) as part of remote play, which is also more urgent given its promise to backers, so I'm backlogging this for now.

smartin015:
Reopening this as the referenced PR has a functional prototype of quest saving.

PR #651, opened 2018-04-17
Quest Saving

smartin015:
An initial (experimental-gated) version of quest saving. 

- It uses local storage to store a fetched quest's XML and the path a user has taken through the quest, along with a timestamp.
- When the user resumes a quest, the node is replayed using the stored path history.
- This probably won't behave 100% correctly if the quest has randomly-generated choices or other random state information (that requires us to store the seed at each step, a change in the QDL code).

Fixes https://github.com/ExpeditionRPG/expedition-app/issues/466


app/actions/SavedQuests.test.tsx
```diff
@@ -0,0 +1,18 @@
+describe('SavedQuest actions', () => {
+  describe('deleteSavedQuest', () => {
+    it('removes both the listing and the data');
```
toddmedema: What would it take to add tests for these functions? The hand effort required to regularly test saving and restoring quest progress is pretty significant, and it's always easier to test something just after you've written it :)

app/components/FeaturedQuests.tsx
```diff
@@ -44,6 +46,13 @@ const FeaturedQuests = (props: FeaturedQuestsProps): JSX.Element => {
           <div className="summary">Explore and play community-written quests.</div>
         </div>
       </Button>
+      {(NODE_ENV === 'dev' || props.settings.experimental) &&
```
toddmedema: Given that NODE_ENV === 'dev' basically implies experimental... what if we set initial experimental state's default based on NODE_ENV === 'dev'? Would get rid of a lot of duplication :)

app/components/base/Card.tsx
```diff
@@ -95,6 +102,7 @@ export default class ExpeditionCard extends React.Component<ExpeditionCardProps,
               {getDevicePlatform() !== 'web' && <MenuItem value="RATE" primaryText="Rate the App"/>}
               <MenuItem value="FEEDBACK" primaryText="Send feedback"/>
               {this.props.inQuest && <MenuItem value="REPORT" primaryText="Report quest"/>}
+              {this.props.inQuest && (NODE_ENV === 'dev' || getStore().getState().settings.experimental) && <MenuItem value="SAVE" primaryText="Save Quest"/>}
```
toddmedema: Would be cleaner / more extensible to fetch settings at the top of render and reference it here

app/actions/SavedQuests.tsx
```diff
@@ -0,0 +1,96 @@
...
+  let details: QuestDetails|null = null;
+    for (let i = 0; i < savedQuests.length; i++) {
+    if (savedQuests[i].details.id === id && savedQuests[i].ts === ts) {
+      details = savedQuests[i].details;
+      break;
+    }
+  }
+
+  if (details === null) {
+    throw new Error('could not load quest details');
```
toddmedema: Capitalize and punctuate.

app/components/base/Card.tsx
smartin015: Done

app/actions/SavedQuests.tsx
smartin015: Done

app/components/FeaturedQuests.tsx
smartin015: Done.

smartin015:
I'm less sold on automatic deletion - that's something done by no other game that I know of.

The randomness issue won't affect more than a small handful of quests, so I don't think that's a blocker in rolling out to beta, just something we need to fix before calling it complete.

app/actions/SavedQuests.test.tsx
smartin015: Done.

PR #652, opened 2018-04-18
add language filter and to quest details

toddmedema:


app/components/Search.tsx
```diff
@@ -169,6 +169,18 @@ class SearchSettingsCard extends React.Component<SearchSettingsCardProps, {}> {
           {rating && <div className="ratingDescription">
             <span>"{this.state.contentrating}" rating means: {rating.summary}</span>
           </div>}
+          <SelectField
+            className="selectfield"
+            floatingLabelText="Language"
+            onChange={(e: any, i: any, v: string) => this.onChange('language', v)}
+            value={this.state.language}
+            style={{color: 'black'}}
```
smartin015: Could we instead move these styles to scss?

app/components/Search.tsx
```diff
@@ -169,6 +169,18 @@ class SearchSettingsCard extends React.Component<SearchSettingsCardProps, {}> {
           {rating && <div className="ratingDescription">
             <span>"{this.state.contentrating}" rating means: {rating.summary}</span>
           </div>}
+          <SelectField
```
smartin015: Thoughts on adding a unit test that confirms the values in the form elements are converted into a valid API request?

app/components/Search.tsx
toddmedema: Ah, yep. Was able to move everything except the underline style, much cleaner!

PR #654, opened 2018-04-19
Add example stateful component test for search settings card

smartin015:


toddmedema:
Also if you could expand it to language as part of this PR since that's merged into master :)

PR #655, opened 2018-04-19
indicate expansion in search results

toddmedema:
Played around with a few different positions, feel like this came out to be the least obtrusive and most extendable as we add more expansions (for example, the top right corner would quickly eat into space for the title, and the bottom left corner would have moved around the time per-result, making it harder to compare quests as you scroll).

screen shot 2018-04-18 at 20 19 56
screen shot 2018-04-18 at 20 19 49

Closes https://github.com/ExpeditionRPG/expedition-app/issues/647


toddmedema:
Closing in favor of https://github.com/ExpeditionRPG/expedition-app/pull/657

PR #658, opened 2018-04-19
lazily check for latest version, notify user if old

toddmedema:
Open to feedback on the text :) Could also, for example, say "New app version available"

Like we have for the quest creator. Realized that this is good to add sooner rather than later since people can download the APK directly / sideload it, and never realize that they're out of date. (or have the app running on an OS that is no longer supported, like iOS 9. Too late to fix that one specifically, but thinking towards the future)

screen shot 2018-04-18 at 21 55 13


app/Main.tsx
```diff
@@ -226,6 +227,28 @@ function handleUrlHash() {
...
+    .then((response: Response) => {
+      return response.json();
+    })
+    .then((response: any) => {
+      const version = response.version;
+      if (version) {
+        const newVersion = version.split('.').map(Number);
+        const oldVersion = getAppVersion().split('.').map(Number);
+        if (newVersion[0] > oldVersion[0] || newVersion[1] > oldVersion[1] || newVersion[2] > oldVersion[2]) {
+          getStore().dispatch(setAnnouncement(true, 'New version available!', 'https://expeditiongame.com/app'));
```
smartin015: Might be good to include a "click here to upgrade" or similar here

app/Main.tsx
```diff
@@ -324,6 +347,9 @@ export function init() {
     } else {
       getStore().dispatch(setDialog('EXPANSION_SELECT'));
     }
+
+    // Things we can wait a while to do / aren't urgent
+    setTimeout(checkVersion, 5 * 1000);
```
smartin015: I feel like after 5 seconds, the user may already be gone from the splash page. Any reason this isn't sooner?

app/Main.tsx
```diff
@@ -226,6 +227,28 @@ function handleUrlHash() {
   }
 }
 
+// If there's a new version of the app available, let the user know
+function checkVersion() {
```
smartin015: Suggested new tests:

inside init:

`it('checks for latest version from github')`

inside `checkVersion`:

`it('displays upgrade announcement if the current app version is different from the hosted version')`

app/Main.tsx
```diff
@@ -226,6 +227,28 @@ function handleUrlHash() {
...
+  fetch('https://raw.githubusercontent.com/ExpeditionRPG/expedition-app/master/package.json')
```
smartin015: What happens in the time between when the new version is submitted and when we release a new android version? Should we be checking an OS-specific location for whether there's an updated app? Might be something cordova already has as a library...

app/Main.tsx
toddmedema: The thinking being that we /really/ don't want to slow down loading for most users, since the vast majority will get auto-updates through the app stores pretty quickly. And that if you use an old app for long enough, you'd eventually see that.

Realizing a problem: we may update master, but it could be several days before Apple approves an update, leading to lots of confused / unhappy users. I guess we could only check on major and minor version to decrease the frequency of that happening?

toddmedema: Update, there's apparently a way to do it with iOS without auth: `http://itunes.apple.com/lookup?bundleId=io.fabricate.expedition`

The Android API requires auth - https://developers.google.com/android-publisher/api-ref/

Given that iOS has the largest publish delay and most problems, I think it would be reasonable to just have two checks: iOS, and Android + Web?

smartin015: Another option would be to add a handler on the API server that does the lookup and returns both versions. WDYT?

For loading slowness... if it's an async fetch, does it really slow down the UI?

app/Main.tsx
toddmedema: Done

app/Main.tsx
toddmedema: Ahhhh, so many good ideas. We can incorporate this into the existing fetch announcement system, which can now also return `{... versions: {ios: 'XXX', android: 'XXXX', web: 'XXXX'}}`, and significantly reduces the amount of new code introduced to the app

toddmedema: I'm thinking we'll want to have the API server cache the calls using something like https://github.com/kwhitley/apicache

We could put a 24 hours cache on announcement and version checks since they aren't updated frequently, but we want the response to be quick to clients / not have them wait on 2-3 external APIs.

app/Main.test.tsx
```diff
@@ -75,6 +75,7 @@ describe('React', () => {
       it('hides android system ui');
       it('pauses music on window pause event');
       it('resumes music on window resume event');
+      it('checks for app update');
```
smartin015: I can haz completed test?

Ditto with checkVersion, actions/Announcement

app/actions/Announcement.tsx
```diff
@@ -13,9 +14,21 @@ export function fetchAnnouncements() {
     .then((response: Response) => {
       return response.json();
     })
-    .then((data: {message: string, link: string}) => {
-      if (data !== null && data.message !== null) {
-        dispatch(setAnnouncement(true, data.message, data.link));
+    .then((data: {message: string, link: string, versions: {android: string, ios: string, web: string}}) => {
```
smartin015: FYI: I added https://github.com/ExpeditionRPG/expedition-qdl/issues/79 to help cut down on these lengthy request object definitions. For now, could you declare it as a `FetchAnnouncementsResponse` interface?

app/actions/Announcement.tsx
```diff
@@ -13,9 +14,21 @@ export function fetchAnnouncements() {
...
+      if (data !== null) {
+        if (data.message !== null && data.message !== '') {
+          dispatch(setAnnouncement(true, data.message, data.link));
+        } else {
+          const version = data.versions[getDevicePlatform()];
+          if (version) {
+            const newVersion = version.split('.').map(Number);
+            const oldVersion = getAppVersion().split('.').map(Number);
+            if (newVersion[0] > oldVersion[0] || newVersion[1] > oldVersion[1] || newVersion[2] > oldVersion[2]) {
```
smartin015: For loop? And ensure version lengths are === 3

app/actions/Announcement.tsx
toddmedema: Done

app/Main.test.tsx
toddmedema: Looks like we're going to need to engineer a good solution for API testing for this one, which is it's own PR in scope. Removed the old test stubs for the old checkVersion stuff, and imported the well-tested `semver` package so that we don't have to worry about testing that functionality.

app/actions/Announcement.tsx
toddmedema: Switched to importing semver instead of reinventing the wheel

PR #660, opened 2018-04-21
Ask to restart after sound disabled; include audio and rp state in sentry errors

toddmedema:


app/components/SettingsContainer.tsx
```diff
@@ -29,6 +30,9 @@ const mapDispatchToProps = (dispatch: Redux.Dispatch<any>, ownProps: any): Setti
   return {
     onAudioChange: (v: boolean) => {
       dispatch(changeSettings({audioEnabled: v}));
+      if (!v) {
+        dispatch(openSnackbar('Please restart the app after disabling sound.'));
```
smartin015: Can we instead auto-disable sound if we detect a problematic/incompatible platform?

toddmedema: From IRL: given that some people also disable sound because they want silence, removing this and replacing with defaulting sound to be based on UNSUPPORTED BROWSERS regex

PR #663, opened 2018-04-24
Confirmation dialog when deleting a save

toddmedema:


app/reducers/Dialog.tsx
```diff
@@ -15,6 +15,7 @@ export function dialog(state: DialogState = initialDialog, action: Redux.Action)
         ...state,
         open: setAction.dialogID,
         message: setAction.message,
+        ts: setAction.ts,
```
smartin015: Is this the ID of the selected saved quest? Why not use `state.saved.selected.ts` instead of making this an attribute of dialog state?

toddmedema: Ok, think I did what you had in mind, but please check the updated commit :)

PR #664, opened 2018-04-24
Moar tests

toddmedema:


app/reducers/User.tsx
```diff
@@ -10,6 +10,14 @@ export const loggedOutUser: UserState = {
   email: '',
 };
 
+export const testLoggedInUser: UserState = {
```
smartin015: I think this would make more sense in User.test.tsx or Checkout.test.tsx, not User.tsx... feels a bit weird to have testing-only code in the file.

toddmedema: Done! Ah yeah, user.test.tsx is an even better spot (didn't want to create a custom user test object in checkout.test because it would be farther away from the actual user code, and will also no doubt be used in other test files)

Issue #571, opened 2018-01-12
Display if quest has already been played in search results and quest details


toddmedema:
Once the API returns IDs and dates of quests played (https://github.com/ExpeditionRPG/expedition-api/issues/62), display a checkmark on the top right of played search results, and "Last Played on XXX" on the quest details page, just above the "Start Quest" button

Need a new `/user/questplays` API route that returns just the id: latest time mapping.

PR #669, opened 2018-04-29
Display to user if they've played a quest before

toddmedema:
Closes https://github.com/ExpeditionRPG/expedition-app/issues/571
screen shot 2018-04-28 at 21 40 49
screen shot 2018-04-28 at 21 39 14
screen shot 2018-04-28 at 21 39 24


app/actions/User.test.tsx
```diff
@@ -2,6 +2,8 @@
 describe('User action', () => {
   describe('silentLogin', () => {
     it('swallows login errors');
+
+    it('loads user quest info on success');
```
smartin015: Populate these?

app/components/Search.tsx
```diff
@@ -273,13 +275,15 @@ function renderDetails(props: SearchProps): JSX.Element {
...
   const ratingAvg = quest.ratingavg || 0;
+  const played = props.user.questPlays[quest.id];
   return (
     <Card title="Quest Details">
       <div className="searchDetails">
         <h2>{quest.title}</h2>
         <div>{quest.summary}</div>
         <div className="author">by {quest.author}</div>
         {(quest.ratingcount && quest.ratingcount >= 1) ? <StarRating readOnly={true} value={+ratingAvg} quantity={quest.ratingcount}/> : ''}
+        {played && <div className="lastPlayed">Last played {Moment(played).format('dddd, MMMM Do')}</div>}
```
smartin015: Specifying the whole date here seems a bit clunky, especially if (as I suspect) folks when searching just want to know "did I play this before".

WDYT about instead a slightly deemphasized (greyed out) style, with an added icon on the top right (a checkmark?) to indicate the quest was previously completed? Then we could show when it was last played on the details card when they actually select the quest.

We could also use `.fromNow()` (https://momentjs.com/docs/#/displaying/fromnow/) to give users more context with less detail.

toddmedema:
Updated UI + added tests:

screen shot 2018-05-06 at 14 42 09
screen shot 2018-05-06 at 14 42 03

app/actions/User.test.tsx
```diff
@@ -1,15 +1,52 @@
...
+
+import {login, silentLogin, UserLoginCallback} from './User'
+import {testLoggedInUser} from '../reducers/User.test'
+import {UserState} from '../reducers/StateTypes'
+import {newMockStore} from '../Testing'
+
+const mockLogin = (callback: UserLoginCallback) => {
+  return callback(testLoggedInUser);
+};
+const callbacks = {
```
smartin015: Have a gander at [jasmine.createSpyObj](https://jasmine.github.io/2.0/introduction.html#section-Spies:_%3Ccode%3EcreateSpyObj%3C/code%3E). Probs easiest to just have this within the unit tests that need it, rather than the spy dance here.

app/actions/Web.tsx
```diff
@@ -39,6 +39,26 @@ export function fetchLocal(url: string) {
...
+        'Content-Type': 'text/plain',
+      },
+    })
+    .then(handleFetchErrors)
+    .then((response: Response) => response.json())
+    .then((quests: UserQuestsType) => {
+      dispatch({type: 'USER_QUESTS', quests} as UserQuestsAction);
+    })
+    .catch((error: Error) => {
+      console.error('Request for quest plays failed', error);
```
smartin015: Should we log this error to more than the console?

app/components/Search.test.tsx
```diff
@@ -84,5 +96,7 @@ describe('Search', () => {
 
   describe('Details', () => {
     it('renders selected quest details');
+    it('shows last played information if it has been played before');
+    it('does not show last played infomation if it does not exist');
```
smartin015: Just stubs? :(

app/reducers/StateTypes.tsx
```diff
@@ -135,14 +135,19 @@ export interface SearchState {
   searching: boolean;
 }
 
-
+export interface UserQuestsType {
```
smartin015: It makes me real sad we're duplicating this interface from [here](https://sourcegraph.com/github.com/ExpeditionRPG/expedition-api/-/blob/app/models/Users.ts#L18). 

Can you instead add this to the QDL schema fileset and import it from there?

app/actions/User.test.tsx
toddmedema: Ah, much nicer!

app/actions/Web.tsx
toddmedema: Eh, nothing breaks in the moment. Once we add a screen for viewing your play history, I think it'd make more sense to show the error on that screen + provide a refresh button.

app/components/Search.test.tsx
toddmedema: Such a tease :P MOAR TESTS! (also, holy cow, once you figure out the general code required to test each new type of setup, adding more tests is a breeze...)

app/reducers/StateTypes.tsx
toddmedema: From IRL: we'll want a separate PR for attempting to import API types into the app as a dev dependency (especially testing that the production bundle size doesn't increase / to make sure it's not actually importing server code)

smartin015: That wasn't really a blocker for adding this to the definitions in expedition-qdl... now we have more redundant interfaces :(

PR #670, opened 2018-04-29
have featured quests also go to quest details screen

toddmedema:
Only one extra tap and gives players a lot more information about the quest (such as duration and when they last played it)


toddmedema:
Test stubs added... need to remember to write full tests as part of the diff so that I don't end up with today's situation of a half-dozen open diffs >.<

PR #673, opened 2018-05-06
log analytics events for quest save and save loads

toddmedema:


app/actions/SavedQuests.tsx
```diff
@@ -41,6 +42,7 @@ export function deleteSavedQuest(id: string, ts: number) {
 }
 
 export function storeSavedQuest(node: ParserNode, details: QuestDetails, ts: number): SavedQuestStoredAction {
+  logEvent('quest_save', { ...details, action: details.title, label: details.id });
```
smartin015: Do we have tests for logEvent? Can we add them now?

Something to the effect of "uses GA when availabe/uses Firebase when available/safely fails when no logging".

I notice in the `logEvent` definition, `argsInput` is an `any` type but is treated as an object. Sounds like a great way to have an unbidden exception to me.

PR #681, opened 2018-05-12
add horror theme

toddmedema:
Desaturates the cards + uses the Kickstarter font for the body. Not 100% sold on the look / open to feedback.

- Tried a few other hand-writing-ish fonts, but they're all comparably difficult to read. Looked at other fonts that tried to be partially-handwritten, but all the ones I found seemed too polished / classy / artificial
- We could also try using it as just a header font, so that it's still apparently but the legibility issues aren't as big a problem. Added a screenshot of that at the end... feel like this actually looks better?

Also fixes a bug whereby exiting out of a quest kept it in the state, which hadn't really been an issue before now but no longer works with theming.

screen shot 2018-05-12 at 14 46 53
screen shot 2018-05-12 at 14 34 12
screen shot 2018-05-12 at 14 34 19
screen shot 2018-05-12 at 14 33 40
screen shot 2018-05-12 at 14 33 35
screen shot 2018-05-12 at 14 33 20

Having it replace header font instead:

screen shot 2018-05-12 at 14 52 50
screen shot 2018-05-12 at 14 52 17
screen shot 2018-05-12 at 14 52 26


app/components/base/Card.tsx
```diff
@@ -86,8 +86,11 @@ export default class ExpeditionCard extends React.Component<ExpeditionCardProps,
       icon = <img id="bgimg" src={'images/' + this.props.icon + '.svg'}></img>;
     }
     const isExperimental = getStore().getState().settings.experimental;
+    const cardTheme = this.props.theme || 'LIGHT';
+    const questTheme = getStore().getState().quest.details.theme || 'base';
+    const classes = ['base_card', 'CARD_THEME_' + cardTheme, 'QUEST_THEME_' + questTheme.toUpperCase()];
```
smartin015: Why not just keep it lowercase? seems like that's what we do for most CSS classes.

app/components/base/Card.tsx
```diff
@@ -86,8 +86,11 @@ export default class ExpeditionCard extends React.Component<ExpeditionCardProps,
       icon = <img id="bgimg" src={'images/' + this.props.icon + '.svg'}></img>;
     }
     const isExperimental = getStore().getState().settings.experimental;
+    const cardTheme = this.props.theme || 'LIGHT';
+    const questTheme = getStore().getState().quest.details.theme || 'base';
```
smartin015: Can we get tests to make sure the theme is applied correctly when there is/isn't a quest?

Issue #682, opened 2018-05-13
Dynamic quest title size, Better summary cutoffs.


smartin015:
The current font size allows for ~17 characters for a quest title in a search card. This leaves a lot of quests overflowing the search card, which looks kind of crappy.

Similarly, we could be a bit smarter about where we cut off the summary text.

PR #685, opened 2018-05-13
Search UI: better text handling

smartin015:
Reduces padding around search results, allows multi-line titles, and makes smarter decisions on where to put ellipses. This should give users a better experience when browsing.

Before:
image

After:
image

Closes https://github.com/ExpeditionRPG/expedition-app/issues/682


package.json
```diff
@@ -21,9 +21,9 @@
     "@types/cheerio": "^0.22.1",
     "@types/enzyme": "^2.7.4",
     "@types/jasmine": "^2.5.43",
-    "@types/material-ui": "^0.20.0",
+    "@types/material-ui": "^0.20.8",
     "@types/promise-polyfill": "^6.0.0",
-    "@types/react": "^15.0.9",
+    "@types/react": "^15.6.15",
```
toddmedema: Can you update the actual react dependency to match?

smartin015: Done.

PR #705, opened 2018-05-24
wrap compositor in container

toddmedema:


app/components/Compositor.tsx
```diff
@@ -131,81 +86,48 @@ export default class Main extends React.Component<MainProps, {}> {
...
+    if (this.props.settings.fontSize === 'SMALL') {
+      containerClass.push('smallFont');
+    } else if (this.props.settings.fontSize === 'LARGE') {
+      containerClass.push('largeFont');
     }
 
     return (
-      <div className={containerClass}>
+      <div className={containerClass.join(' ')}>
         <Provider store={getStore()}>
```
smartin015: The Provider should be moved out to `Main.tsx` to align with typical react-redux ways (see https://redux.js.org/basics/usage-with-react#passing-the-store)

app/components/views/quest/cardtemplates/Template.tsx
```diff
@@ -57,8 +55,9 @@ export function getCardTemplateTheme(card: CardState): CardThemeType {
     case 'NO_TIMER':
     case 'MID_COMBAT_ROLEPLAY':
       return 'dark';
+    // case 'ROLEPLAY':
```
smartin015: Why not un-comment and have it fall through?

app/components/Compositor.tsx
```diff
@@ -17,84 +31,31 @@ import SettingsContainer from './views/SettingsContainer'
...
-    if (!state.card || (this.state && newTS === this.state.ts && newKey === this.state.key)) {
-      return this.state;
-    }
+export default class Compositor extends React.Component<CompositorProps, {}> {
 
-    let card: JSX.Element;
-    let theme: CardThemeType = 'light';
-    switch(state.card.name) {
+  render() {
+    let card = <SplashScreenContainer/>;
```
smartin015: Why initialize it with something when we have a strict `switch` statement directly after?

app/Main.tsx
```diff
@@ -327,15 +327,15 @@ export function init() {
...
+  const CompositorContainer = require('./components/CompositorContainer').default;
   const base = getDocument().getElementById('react-app');
   if (!base) {
     throw new Error('Could not find react-app element');
   }
   ReactDOM.unmountComponentAtNode(base);
   ReactDOM.render(
     <MuiThemeProvider muiTheme={getMuiTheme(theme)}>
-      <Compositor/>
+      <CompositorContainer store={getStore()}/>
```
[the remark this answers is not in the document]
toddmedema: Ah, good call!

app/components/Compositor.tsx
toddmedema: Done

app/components/views/quest/cardtemplates/Template.tsx
toddmedema: Linter won't let me

app/components/Compositor.tsx
toddmedema: Gives us free type initialization, primarily. But reduced this to `<span/>` to be clearer about that

app/components/Compositor.test.tsx
```diff
@@ -1,5 +1,3 @@
-describe('Main', () => {
-  it('does not trigger animation if state update to same card');
-
-  it('intercepts and publishes touch/move events');
+describe('Compositor', () => {
+  it('Renders the provided card');
 });
```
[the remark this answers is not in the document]
toddmedema: Added. Not sure what else to test here besides that it renders a provided card. Seems low value / high maintenance to write tests for each of the switch cases.

app/components/Compositor.test.tsx
```diff
@@ -1,5 +1,3 @@
-describe('Main', () => {
-  it('does not trigger animation if state update to same card');
-
-  it('intercepts and publishes touch/move events');
+describe('Compositor', () => {
+  it('Renders the provided card');
```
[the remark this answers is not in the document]
toddmedema: Done

PR #706, opened 2018-05-24
streamline audio component

toddmedema:
- Remove this.loaded, use redux state value instead
- Dedupe private attributes of audio against props for the component
- AudioLoadingType should be just string types


app/components/base/Audio.tsx
```diff
@@ -136,22 +128,26 @@ export default class Audio extends React.Component<AudioProps, {}> {
 
     const newIntensity = Math.round(Math.min(MUSIC_INTENSITY_MAX, Math.max(0, nextProps.audio.intensity)));
 
-    if (this.enabled !== nextProps.enabled) {
+    if (this.props.enabled !== nextProps.enabled) {
       if (nextProps.enabled) {
-        this.enabled = true;
         this.loadFiles();
         this.playAtIntensity(this.intensity);
```
smartin015: Why is `resumeMusic` required here? Does playAtIntensity not actually play the music any more? If so, suggest renaming to `setIntensity` or similar.

toddmedema: Fair point, led to a redundant call. Still have to set `this.paused = false`, though, since when you disable it, it pauses the music.

Issue #450, opened 2017-09-10
Make 0 search results UI more fun


toddmedema:
*Migrated to expeditionrpg/expedition#219 by [spacetelescope/github-issues-import](https://github.com/spacetelescope/github-issues-import)*

Earns goodwill; also a chance to provide a bit of instruction, or a handy shortcut for broadening results

toddmedema:
Backlogging since it didn't make it into this bug sprint

ezittel:
#709 added

ezittel:
![image](https://user-images.githubusercontent.com/16277302/40544333-e3406744-5ff5-11e8-80a8-7a7c2feb25cb.png)

PR #709, opened 2018-05-25
Make 0 results more fun (#450)

ezittel:
It's probably not more fun, but the link is added.

### DIFF
--- a/app/components/base/AudioContainer.test.tsx
+++ b/app/components/base/AudioContainer.test.tsx
@@ -1,3 +1,0 @@
-describe('AudioContainer', () => {
-  // Currently simple enough
-});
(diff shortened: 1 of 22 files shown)


toddmedema:
Awesome! Can you attach a screenshot of the new UI?

ezittel:
Looks something like:
image

toddmedema:
Looks good to me! Note that for more complicated UI features, we like to use screenshots for both desktop and mobile (via the Chrome mobile emulator in dev tools) since most of our players use the mobile experience. But this is a straightforward addition, so we can skip that this time :)

ezittel:
Okay, good to know.  Although I'm not sure I like the mobile results:

ezittel:
I cleaned this up, shortened the text a little so it looks good on 
mobile as well, then attached both screenshots to a new PR.

Sorry for the confusion.

PR #717, opened 2018-05-30
Quest Awarded and Official

ezittel:
Added all necessary changes for awarded/official icons style changes and text.

### DIFF
--- a/app/Style.scss
+++ b/app/Style.scss
@@ -1118,6 +1118,7 @@ span.line {
       font-size: fontsize(interactive);
       margin-bottom: size(small);
       top: 0;
+      width: 85%;
       white-space: nowrap;
       overflow: hidden;
     }
@@ -1195,7 +1196,8 @@ span.line {
   }
   .indicator_spacer{
     position:relative;
-    top: -(size(tiny));
+    left: size(tiny);
+    top: -(size(small));
   }
   .questButtonWithIcon {
     .inline_icon {
--- a/app/Style.scss
+++ b/app/Style.scss
@@ -34,7 +34,7 @@
   transition: transform 300ms ease-in-out;
 }
 .NEXT-leave {
-  position:absolute;
+  position: absolute;
   transform: translate(0%);
 }
 .NEXT-leave.NEXT-leave-active {
@@ -50,7 +50,7 @@
   transition: transform 300ms ease-in-out;
 }
 .PREV-leave {
-  position:absolute;
+  position: absolute;
   transform: translate(0%);
 }
 .PREV-leave.PREV-leave-active {
@@ -167,7 +167,7 @@ $white_on_red_text_shadow: 0 1px 2px rgba(0,0,0,0.3) !important;
 
 body {
   background-color: $bg_inactive;
-  overflow:hidden;
+  overflow: hidden;
   width: 100%;
   height: 100%;
   margin: 0;
@@ -192,7 +192,7 @@ body {
 }
 
 .remote_footer {
-  position:absolute;
+  position: absolute;
   bottom: 0;
   height: size(remotefooter);
   width: 100%;
@@ -227,7 +227,7 @@ body {
   width: 100%;
   height: 100%;
   background: black;
-  position:fixed;
+  position: fixed;
   z-index: 1000;
 
   > div {
@@ -808,7 +808,7 @@ span.line {
   }
 
   .base_multi_touch_remote_affector {
-    position:absolute;
+    position: absolute;
     width: 100%;
     height: 100%;
   }
@@ -1118,7 +1118,7 @@ span.line {
       font-size: fontsize(interactive);
       margin-bottom: size(small);
       top: 0;
-      overflow:hidden;
+      overflow: hidden;
     }
     .summary {
       display: flex;
@@ -1181,23 +1181,17 @@ span.line {
       }
     }
     .searchResultsTitleTable {
-      width:100%;
+      width: 100%;
       .leftcell{
-          //border: 1px solid red;
-          overflow:hidden;
-        th {
+          overflow: hidden;
           text-align: left;
-        }
       }
       .rightcell{
-        display:block;
+        display: block;
         white-space: nowrap;
-        overflow:hidden;
+        overflow: hidden;
         text-align: right;
-        th {
-          width:auto;
-          text-align: right;
-        }
+        width: auto;
       }
 
     }
@@ -1210,7 +1204,7 @@ span.line {
         width: fontsize(interactive) !important;
       }
       .indicator_spacer{
-      position:relative;
+      position: relative;
       top: -(size(tiny));
       }
     }
@@ -1231,12 +1225,12 @@ span.line {
       font-style: italic;
     }
     .inline_icon {
-      position:relative;
+      position: relative;
       top: size(base);
       padding-bottom: size(base);
     }
     .indicators {
-      position:relative;
+      position: relative;
       top: -(size(base));
       opacity: 0.65;
       padding-top: size(small);
@@ -1255,7 +1249,7 @@ span.line {
     margin-top: size(large);
     .searchDetailsTable {
       width: 100%;
-      table-layout:fixed;
+      table-layout: fixed;
       th {
         text-align: left;
         max-width: 50%;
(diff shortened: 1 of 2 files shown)


toddmedema:
In the screenshot you attached with the titles being fixed, the window was too wide to actually observe the title overlap behavior - could you take another screenshot on a narrower window so that I can see what that looks like?

app/components/views/Search.tsx
```diff
@@ -328,6 +332,8 @@ export function renderDetails(props: SearchDetailsProps): JSX.Element {
         {(quest.ratingcount && quest.ratingcount >= 1) ? <StarRating readOnly={true} value={+ratingAvg} quantity={quest.ratingcount}/> : ''}
         <div className="indicators">
           {props.lastPlayed && <div className="lastPlayed"><DoneIcon className="inline_icon" /> Last played {Moment(props.lastPlayed).fromNow()}</div>}
+          {quest.official && <span> <img className="inline_icon" src="images/compass_small.svg"/> Official Quest!</span>}
+          {quest.awarded && <div className="awarded"><StarsIcon className="starsIcon" /> {quest.awarded}</div>}
```
toddmedema: Note that this should have the `inline_icon` class like the lastPlayed icon above to get the fully correct styling; we also don't need to apply a `starsIcon` class, since this icon shouldn't have it's own specific styling - the styling should be generic across all icons / not care about what icon it is (as it is with `inline_icon`)

app/components/views/Search.tsx
```diff
@@ -271,7 +272,10 @@ export function renderResult(props: SearchResultProps): JSX.Element {
         <span className="expansions">
           {quest.expansionhorror && <img className="inline_icon" src="images/horror_small.svg"></img>}
         </span>
-        <div className="indicators">{props.lastPlayed && <DoneIcon className="questPlayedIcon" />}</div>
+        <div className="indicators">{props.lastPlayed && <DoneIcon className="questPlayedIcon" />}
+          {quest.official !== undefined && quest.official && <span className="indicator_spacer"><img className="inline_icon" src="images/compass_small.svg"/></span>}
+          {quest.awarded && <StarsIcon className="starsIcon" />}
```
toddmedema: (see note below about having this be `inline_icon` instead of `questPlayedIcon` and `starsIcon`)

app/Style.scss
```diff
@@ -1120,6 +1118,7 @@ span.line {
       font-size: fontsize(interactive);
       margin-bottom: size(small);
       top: 0;
+      width: 85%;
```
toddmedema: Hmm - this isn't ideal since it cuts things off a fixed amount regardless of the number of icons - cutting off too much when there's no icons, and not enough when there are more icons. 

If the float part didn't work, you could try this:
- wrap the .title div in a table, and have the title be one cell
- move the .indicators div into a cell in the same row and remove the position styling on `.indicators`

Getting things to appear just right in HTML / CSS is one of the harder / more annoying parts of web development. If you ever get really frustrated at this, let me know and I can take over that part of the diff - but, if you're interested in learning, learning by doing is definitely the best way to go about it!

app/components/views/Search.tsx
ezittel: Details page:  When I changed the className I get the first picture.  I made css changes and the 2nd screenshot shows the fix.
Good?
image

image

toddmedema: Can you push your changes so I can see how you did that in the code? :)

app/Style.scss
ezittel: Yeah float doesn't work on title and since there is position in the .indicators float is ignored.  I'll try the table approach, I do this trick all the time.

ezittel:
I played with the CSS quite a bit and am closer to having something - will have you review tomorrow.

ezittel:
this is worth a quick review at this point.

app/Style.scss
```diff
@@ -1183,15 +1180,39 @@ span.line {
         }
       }
     }
+    .searchResultsTitleTable {
+      width:100%;
```
toddmedema: Style nit, all css properties should have a space after the colon :)

app/Style.scss
```diff
@@ -1183,15 +1180,39 @@ span.line {
...
+      .leftcell{
+          //border: 1px solid red;
+          overflow:hidden;
+        th {
```
toddmedema: note that this left align won't do anything since nested properties select elements /below/ the original element in the tree. Since the `tr` has the class `leftcell` directly, you can just delete the `tr` part of this and have text-align apply direct to leftcell

app/Style.scss
```diff
@@ -1183,15 +1180,39 @@ span.line {
...
+          text-align: left;
+        }
+      }
+      .rightcell{
+        display:block;
+        white-space: nowrap;
+        overflow:hidden;
+        text-align: right;
+        th {
```
toddmedema: (same note here about being able to get rid of the nested th selector)

app/Style.scss
```diff
@@ -1183,15 +1180,39 @@ span.line {
...
+          //border: 1px solid red;
+          overflow:hidden;
+        th {
+          text-align: left;
+        }
+      }
+      .rightcell{
+        display:block;
+        white-space: nowrap;
+        overflow:hidden;
```
toddmedema: curious why overflow: hidden is needed?

app/Style.scss
```diff
@@ -1183,15 +1180,39 @@ span.line {
...
-      right: size(base);
+      position: relative;
+      margin-top: size(tiny);
       opacity: 0.65;
       svg {
         height: fontsize(interactive) !important;
         width: fontsize(interactive) !important;
       }
+      .indicator_spacer{
+      position:relative;
```
toddmedema: nit: indentation :)

app/components/views/Search.tsx
```diff
@@ -252,11 +253,23 @@ export function renderResult(props: SearchResultProps): JSX.Element {
   return (
     <Button key={props.index} onTouchTap={() => props.onQuest(quest)} remoteID={'quest-'+props.index}>
       <div className={classes.join(' ')}>
-        <div className="title">
-          <Truncate lines={2}>
-            {quest.title}
-          </Truncate>
-        </div>
+        <table className="searchResultsTitleTable">
+          <tr>
```
toddmedema: I think you might see a warning in the console about `<table>` needing a `<tbody>` to wrap the `<tr>`'s

app/components/views/Search.tsx
```diff
@@ -252,11 +253,23 @@ export function renderResult(props: SearchResultProps): JSX.Element {
...
+            <th className="leftcell">
+              <div className="title">
+                <Truncate lines={2}>
+                  {quest.title}
+                </Truncate>
+              </div>
+            </th>
+            <th className="rightcell">
+              <div className="indicators">{props.lastPlayed && <DoneIcon className="inline_icon" />}
```
toddmedema: Thinking you could probably get rid of this `indicators` div and just make the `th` of class `indicators` - and same thing with the title, make the `th` of class `title` and remove the extra div / adjust the css to match the change.

app/components/views/Search.tsx
ezittel: no but can throw one in

app/Style.scss
ezittel: The scrollbar kept showing up for some reason, of course that's when I had it in mobile mode and dragging the size around to see what would happen - probably not a good use case, but the scrollbar doesn't show up anymore.

app/Style.scss
ezittel: fixed

app/Style.scss
ezittel: nod nod

app/components/views/Search.tsx
ezittel: yeah I'm scared to touch it now.  see screen shots

ezittel: image
image

ezittel: oh this indicators adjustment is the things that keeps that compass icon aligned vertically.  I'm not getting what you mean by this.

> Thinking you could probably get rid of this indicators div and just make the th of class indicators - and same thing with the title, make the th of class title and remove the extra div / adjust the css to match the change.

ezittel:
Please review.

toddmedema:
Noting here - not sure why CI didn't run tests against your commits, but looks like it broke a few tests, so fixed that and added more tests to check against the new functionality here - https://github.com/ExpeditionRPG/expedition-app/commit/0faa4ddb60affaf18dbaa1c1a0ee8a5e1ba311d7

PR #722, opened 2018-06-03
Cleanup login methods & update basically everything

smartin015:
**Changes:**

- Cleaned up login flow (promisifies most methods and makes the callee code much cleaner)
- Remove inject-tap-event-plugin which is unneeded with react 16 (use onClick instead)
- Remove redux-thunk and add in shadow of react dispatch() method (which is what we were using it for anyways)
- Migrate to React 16, Typescript 2.8.3, Material UI 1.0, and Redux 4. Bring all additional dependent libs up to a compatible version. Note that newer versions of typescript are incompatible with React 16.
- Streamlined our webpack configs a bit (awesome-typescript-loader can actually be pointed to use Babel on its own, no need for post-loading).

**This PR comes with some regressions:**

- CSS styles are mostly ok, but the search UI needs some TLC and some icon sizing may have changed a bit. We should switch to using CSS-in-JS, but I'd much prefer that be part of a separate PR/effort.
- I commented out several of the UI tests which failed to build... pretty sure the tests are wrong here, not the UI itself. We should fix these afterwards.
- Remote-generated tap events will not produce ripples on buttons.


toddmedema:
As far as the regressions:
- would be great if we could get the CSS to be comparable to where it was before, or at least actively not bad, otherwise master will be broken
- tests and remote ripples are both things I'm fine with us fixing in a follow-up

app/actions/Checkout.tsx
```diff
@@ -15,18 +14,11 @@ export function checkoutSetState(delta: Partial<CheckoutState>) {
...
-        logEvent('to_checkout', {value: amount});
-        dispatch(toCard({name: 'CHECKOUT', phase: 'ENTRY'}));
-      }}));
-    } else {
-      logEvent('to_checkout', {value: amount});
-      dispatch(toCard({name: 'CHECKOUT', phase: 'ENTRY'}));
-    }
-  }
+    logEvent('to_checkout', {value: amount});
+    dispatch(toCard({name: 'CHECKOUT', phase: 'ENTRY'}));
```
toddmedema: Why'd we get rid of the login logic before tipping?

app/components/Compositor.test.tsx
```diff
@@ -1,17 +1,18 @@
 import * as React from 'react'
-import {render} from 'enzyme'
+import {configure, render} from 'enzyme'
+import * as Adapter from 'enzyme-adapter-react-16'
+configure({ adapter: new Adapter() });
+
 import {Provider} from 'react-redux'
-import getMuiTheme from 'material-ui/styles/getMuiTheme'
+//import createMuiTheme from '@material-ui/core/styles/createMuiTheme'
```
toddmedema: Should we rm the commented out stuff in this file? (and other test files that have this removed)

app/components/Compositor.tsx
```diff
@@ -102,28 +108,34 @@ export default class Compositor extends React.Component<CompositorProps, {}> {
...
-              transitionLeaveTimeout={CARD_TRANSITION_ANIMATION_MS}>
-            <div className={'base_main' + ((this.props.remotePlay && this.props.remotePlay.session) ? ' has_footer' : '')} key={this.props.card.key}>
-              {card}
-            </div>
-          </ReactCSSTransitionGroup>
+          <TransitionGroup
+            childFactory={(child) => React.cloneElement(
+              child, {classNames: this.props.transition}
+            )}>
+            <CSSTransition
```
toddmedema: Oh boy, we had to change this? I'll be interested to see if this upgrade improves mobile transition performance - *crosses fingers*

app/components/base/StarRating.tsx
```diff
@@ -27,21 +27,24 @@ export default class StarRating extends React.Component<StarRatingProps, {}> {
         classes.push('editable');
       }
 
+      // TODO Fix icon colors: https://material-ui.com/style/icons/
```
toddmedema: Tagging this as a regression we should ideally fix before merging into master

app/actions/Checkout.tsx
smartin015: It's duplicated in QuestEndContainer's onTip() method.

app/components/Compositor.test.tsx
smartin015: Fair point... removed what I could find (upcoming commit)

app/components/Compositor.tsx
smartin015: Yep. I'm honestly surprised it was as drop-in as it was.

toddmedema:
Search styles good enough for now; still an odd bug where search props not established in the props when you go to the filter page have their labels return to the select area even if you pick a value (instead of staying above the select), but I'm fine with dealing with that later

screen shot 2018-06-03 at 14 56 06

toddmedema:
Oi, looks like the post-tip flow crashes the app. So close...

app/components/base/StarRating.tsx
smartin015: Fixed in latest commit.

PR #723, opened 2018-06-06
Code Review Fixes

smartin015:
Fixed the following cleanup items from the code review:

- Move local storage logic to its own file (Storage.tsx?) And add test
- openSnackbar should take string\|error instead of separate errorMessage
- Un-functionify postTimerReturn
- Move "loading delay milliseconds" constants from various places to Constants.tsx
- Dedupe onTip
- Lowercase animation classes
- Dedupe some common dialog stuff (e.g. TextFields for reports)


app/Constants.tsx
```diff
@@ -47,6 +47,11 @@ export const URLS = {
   QUEST_CREATOR: 'https://quests.expeditiongame.com/?utm_source=app',
 };
 
+export const INIT_DELAY = {
+  SILENT_LOGIN_MILLIS: 2000,
```
toddmedema: Thoughts on bumping this up (down?) to 1000 or even 500? If you know what you're doing in the app, it's possible to get all the way to search before login has triggered, and it's a low-impact request

PR #728, opened 2018-06-07
Fixes

toddmedema:
refactor and fix checkout
switch from pre-commit to pre-push
fix audio controls
live-calculate timer duration based on up to date settings to allow for changing adventurers setting mid-combat


app/components/views/Checkout.tsx
```diff
@@ -77,37 +78,50 @@ export default class Checkout extends React.Component<CheckoutProps, {}> {
...
+            {this.props.checkout.processing && <div className="centralMessage">Processing payment, one moment...</div>}
+          </form>
+        </div>
+      </Card>
+    );
+  }
+}
+
+
+class CheckoutThankYou extends React.Component<CheckoutProps, {}> {
```
smartin015: Probably best if this were just a function (like [these](https://github.com/ExpeditionRPG/expedition-app/blob/master/app/components/views/quest/cardtemplates/combat/Combat.tsx#L59))

app/components/views/quest/cardtemplates/combat/Actions.tsx
```diff
@@ -19,7 +19,8 @@ import {getStore} from '../../../../../Store'
 const cheerio: any = require('cheerio');
 
 // Number of adventurers (such as for damage)
-function numLocalAndMultiplayerAdventurers(settings: SettingsType, rp: MultiplayerState): number {
+// TODO test
```
smartin015: Move this to the test file, add stubs (here & below), and remove the comment on L21 (I don't think it adds value)

app/components/views/quest/cardtemplates/combat/Actions.tsx
```diff
@@ -19,7 +19,8 @@ import {getStore} from '../../../../../Store'
...
+export function _numLocalAndMultiplayerAdventurers(settings: SettingsType, rp: MultiplayerState): number {
```
smartin015: `export` implies "used elsewhere", leading underscore implies local. Which one's correct?

app/components/views/quest/cardtemplates/combat/Actions.tsx
```diff
@@ -37,7 +38,8 @@ function numLocalAndMultiplayerAdventurers(settings: SettingsType, rp: Multiplay
 }
 
 // Number of fingers
-function numLocalAndMultiplayerers(settings: SettingsType, rp: MultiplayerState): number {
+// TODO test
+export function _numLocalAndMultiplayerers(settings: SettingsType, rp?: MultiplayerState): number {
```
smartin015: `Multiplayerers`?! What is this thing?

app/components/views/quest/cardtemplates/combat/Combat.tsx
```diff
@@ -440,7 +440,7 @@ function renderTimerCard(props: CombatProps): JSX.Element {
       tertiaryText={instruction}
       icon={enemyClass}
       numPlayers={(props.settings.multitouch && props.settings.numPlayers > 1) ? props.numAliveAdventurers : 1}
-      roundTimeTotalMillis={props.combat.roundTimeMillis}
```
smartin015: Do we not need `roundTimeMillis` in CombatState any more?

app/components/views/quest/cardtemplates/combat/Actions.tsx
toddmedema: lol, done

app/components/views/quest/cardtemplates/combat/Combat.tsx
toddmedema: Nope, it's a calculated parameter. If we wanted to get really fancy here, we might do something like a selector as the official way to handle derived data - but that doesn't seem worth getting into right now. https://redux.js.org/recipes/computing-derived-data

app/components/views/quest/cardtemplates/combat/Actions.tsx
toddmedema: Only exported for tests. Open to other ways of handling that :)

app/components/views/quest/cardtemplates/combat/Actions.tsx
toddmedema: Done

app/components/views/Checkout.tsx
toddmedema: Done

app/components/views/quest/cardtemplates/combat/Combat.tsx
smartin015: If it's calculated, let's remove it from `CombatState` defined in cardtemplates/combat/Types.tsx.

app/components/views/quest/cardtemplates/combat/Actions.tsx
smartin015: As discussed, let's privatize these and instead test them indirectly through the interface.

app/components/views/quest/cardtemplates/combat/Combat.tsx
toddmedema: It is removed from there already

PR #730, opened 2018-06-08
fix login bugs

toddmedema:
- cleans up promises indentations so that the `then`s are all on the same level as the original promise (I've come to prefer that style, but open to thoughts)
- fixes bug where user info wasn't getting returned from a login, causing things like reporting a quest to fail if you weren't logged in -> auth -> user not defined
- fixes a bug where going backwards / `RETURN`ing would drop user from state
- confirmed working on both Android and iOS


toddmedema:
Note, the `null` -> `undefined` fix is blocking another contributor, so I'm going to merge this in now and we can do a post-push review on it.

PR #732, opened 2018-06-09
update user quests in state after play complete

toddmedema:
Almost to the point that we can show users a page with past quests they've played (their "questing history")

Also, gotta love deepmerge module - awesome functionality with no deps, only adds 500B of code - https://www.npmjs.com/package/deepmerge


app/actions/Web.tsx
```diff
@@ -91,10 +91,12 @@ function loadQuestXML(a: {details: QuestDetails, questNode: Cheerio, ctx: Templa
 
 export function logQuestPlay(a: {phase: 'start'|'end'}) {
```
smartin015: I'd make this a redux action, since we would have more principled access to the store and dispatch function.

toddmedema: Done

PR #734, opened 2018-06-10
First wack at writing a test.

ezittel:
Pretty rudimentary, but gotta start somewhere.

### DIFF
--- a/app/components/views/Search.test.tsx
+++ b/app/components/views/Search.test.tsx
@@ -143,6 +143,15 @@ describe('Search', () => {
       expect(wrapper.html()).toContain(quest.genre);
       expect(wrapper.html()).toContain(quest.summary);
       expect(wrapper.html()).toContain(quest.author);
+      expect(wrapper.html()).toContain(quest.official);
+    });
+
+    it('does not render unselected quest details', () => {
+      const quest = FEATURED_QUESTS.filter((el) => el.title === 'Learning to Adventure')[0];
+      const {props, wrapper} = setup(quest.title);
+      expect(wrapper.html()).not.toContain(quest.expansionhorror);
+      expect(wrapper.html()).not.toContain(quest.requirespenpaper);
+      expect(wrapper.html()).not.toContain(quest.awarded);
     });
 
     it('shows last played information if it has been played before', () => {
--- a/app/components/views/Search.test.tsx
+++ b/app/components/views/Search.test.tsx
@@ -75,10 +75,9 @@ describe('Search', () => {
       return {props, wrapper};
     }
 
-    it('displays no expansion icons when quest has no expansions', () => {
+    it('displays no horror icon when a quest does not use the Horror expansion', () => {
       const {props, wrapper} = setup('Learning to Adventure');
       expect(wrapper.html()).not.toContain('horror');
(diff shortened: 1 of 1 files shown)


ezittel:
Is there a way to run just one test class from the command line?

app/components/views/Search.test.tsx
```diff
@@ -141,6 +143,15 @@ describe('Search', () => {
       expect(wrapper.html()).toContain(quest.genre);
       expect(wrapper.html()).toContain(quest.summary);
       expect(wrapper.html()).toContain(quest.author);
+      expect(wrapper.html()).toContain(quest.official);
+    });
+
+    it('does not render unselected quest details', () => {
```
toddmedema: I would actually merge this in with the "rendered selected quest details" above, since this is all about rendering the selected quest appropriately. Note that I don't think this will detect these properly, since these values are all false or undefined, so it's checking to see if the html does not contain false / undefined rather than not containing the user-facing indication of those attributes

app/components/views/Search.test.tsx
```diff
@@ -77,6 +78,7 @@ describe('Search', () => {
     it('displays no expansion icons when quest has no expansions', () => {
       const {props, wrapper} = setup('Learning to Adventure');
       expect(wrapper.html()).not.toContain('horror');
+      expect(wrapper.html()).not.toContain('book');
```
toddmedema: Thinking it would make more sense to split this out into two separate 'displays pen and paper requirement when set / does not display pen and paper requirement when set' test cases to keep the style here. Plus, it's already set up to make it easy to inject specific quest attributes:
`const {props, wrapper} = setup('Learning to Adventure', {requirespenpaper: true});`

toddmedema:
These are great additions - a perfect place to start! 

I haven't tried running a specific test with this setup before, but I imagine it would work something like this - https://stackoverflow.com/a/29151264 - since our `npm run test` uses Karma

app/components/views/Search.test.tsx
ezittel: okay so is there a correct way to do this? I guess I'll have to assert on null somehow for wrapper.html() and 'horror'?

app/components/views/Search.test.tsx
ezittel: I"ll look closer at setup to see what kind of options I've got.
Thanks for the continued feedback, it means a lot!

app/components/views/Search.test.tsx
```diff
@@ -74,7 +75,7 @@ describe('Search', () => {
       return {props, wrapper};
     }
 
-    it('displays no expansion icons when quest has no expansions', () => {
+    it('displays no horror icon when a quest does not use the Horror expansion', () => {
```
toddmedema: Perfect! Can you create a test similar to this, but extends `setup()` with quest overrides `{requirespenpaper: true}` and confirms that the icon does appear - note that since quest overrides is the third param, you'll have to pass an empty object as the second param to setup

toddmedema:
Don't want your work here to get lost! Can you migrate these changes to the new monorepo then close this PR?

ezittel:
migrated and closed.

ezittel:
https://github.com/ExpeditionRPG/expedition/pull/6

PR #737, opened 2018-06-13
First pass attempt at Decision logic (dev-gated) plus migration of remoteID -> ID

smartin015:


app/Constants.tsx
```diff
@@ -1,6 +1,6 @@
 import {QuestDetails} from './reducers/QuestTypes'
 
-export const NODE_ENV = (process && process.env && process.env.NODE_ENV) || 'dev';
+export const NODE_ENV = (process && process.env && process.env.NODE_ENV) || 'prod';
```
toddmedema: We should also update the hard coded defaults in webpack - https://github.com/ExpeditionRPG/expedition-app/blob/master/webpack.dist.config.js#L34

app/Style.scss
```diff
@@ -1296,6 +1302,17 @@ span.line {
...
   }
 
+  .multiplayer_ripple {
+    height: '100%';
+    width: '100%';
+    position: 'absolute';
+    top: 0;
+    left: 0;
+    overflow: 'hidden';
+    pointerEvents: 'none';
```
toddmedema: if this is CSS, these should be pointer-events and z-index, no?

app/components/views/Tools.tsx
```diff
@@ -39,13 +39,13 @@ const Tools = (props: ToolsProps): JSX.Element => {
...
       }
-      {!props.settings.simulator && <Button remoteID="1" id="selectQuestCreator" onClick={() => props.onQuestCreatorSelect()}>
+      {!props.settings.simulator && <Button id="selectQuestCreator" onClick={() => props.onQuestCreatorSelect()}>
         <div className="questButtonWithIcon">
           <div className="title">Quest Creator</div>
           <div className="summary">Write your own quests and share them with the world.</div>
         </div>
       </Button>}
-      {!props.settings.simulator && <Button remoteID="2" onClick={() => props.onPrivateQuestsSelect(props.settings, props.user)}>
+      {!props.settings.simulator && <Button id="2" onClick={() => props.onPrivateQuestsSelect(props.settings, props.user)}>
```
toddmedema: Can this ID be something more useful like #privateQuests?

app/components/views/quest/cardtemplates/combat/Actions.tsx
```diff
@@ -72,12 +73,47 @@ export function generateCombatTemplate(settings: SettingsType, custom: boolean,
...
 
+interface ToDecisionCardArgs {
+  node?: ParserNode;
+  phase: DecisionPhase;
+  numOutcomes?: number;
+  settings?: SettingsType;
+}
+export const toDecisionCard = remoteify(function toDecisionCard(a: ToDecisionCardArgs, dispatch: Redux.Dispatch<any>, getState: () => AppStateWithHistory): ToDecisionCardArgs {
+  if (!a.node) {
+    a.node = getState().quest.node;
```
toddmedema: Isn't this logic comparable to the more succinct `a.node = a.node || getState().quest.node`? (same below with settings)

app/components/views/quest/cardtemplates/decision/Actions.tsx
```diff
@@ -0,0 +1,161 @@
...
+function generateDecisionTemplate(): DecisionState {
+  return {
+    scenario: EMPTY_SCENARIO,
+    numAttempts: 0,
+    outcomes: [],
+    choice: EMPTY_DECISION,
+  };
+}
+
+export function roundTimeMillis(settings: SettingsType, rp?: MultiplayerState) {
```
toddmedema: Also TODO dedupe or clarify how this is different than roundTimeMillis in combat (even just calling it skillTimeMillis and using a different constant would be sufficient)

app/components/views/quest/cardtemplates/decision/Types.tsx
```diff
@@ -0,0 +1,57 @@
+
```
toddmedema: Lots spaces?

package.json
```diff
@@ -71,6 +71,7 @@
   "dependencies": {
     "@material-ui/core": "^1.0.0-rc.0",
     "@material-ui/icons": "^1.1.0",
+    "@types/classnames": "^2.2.4",
```
toddmedema: Types for the package added, but not the package?

app/Constants.tsx
smartin015: Reverted after deciding it's not worth fixing in this PR.

app/Style.scss
smartin015: Done.

app/components/views/Tools.tsx
smartin015: Done.

app/components/views/quest/cardtemplates/combat/Actions.tsx
smartin015: It is indeed, but it's also a standard style used in the Combat actions and in other actions files. I think we should migrate these wholesale, later.

app/components/views/quest/cardtemplates/decision/Actions.tsx
smartin015: Done.

app/components/views/quest/cardtemplates/decision/Types.tsx
smartin015: Done.

package.json
smartin015: Added

app/Style.scss
smartin015: Done.

